PR #20, opened 2019-04-02
feat(dialog): add dividers on header/footer on content scroll

matmkian:


src/components/dialog/angularjs/dialog_directive.js
```diff
@@ -193,6 +160,40 @@ function DialogController(
         }, _TRANSITION_DURATION);
     }
 
+    /**
+     * Create dialog content observer.
```
mcastagnetti: Maybe you should be a little bit more exhaustive about what you do with those observers

PR #40, opened 2019-04-15
feat(userblock): created userblock component

GuillaumeNachury:
Create new component to display user details

### DIFF
--- a/src/components/user-block/react/UserBlock.tsx
+++ b/src/components/user-block/react/UserBlock.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import classNames from 'classnames';
 
-import { Orientations, Theme, Themes } from 'LumX/components';
+import { Orientation, Orientations, Theme, Themes } from 'LumX/components';
 
 import { COMPONENT_PREFIX } from 'LumX/core/react/constants';
 
@@ -38,7 +38,7 @@ interface IUserBlockProps extends IGenericProps {
     /* User name. */
     name?: string;
     /* Orientation. */
-    orientation?: Orientations;
+    orientation?: Orientation;
     /* Size. */
     size?: Size;
     /* Theme. */
--- a/src/components/user-block/react/UserBlock.tsx
+++ b/src/components/user-block/react/UserBlock.tsx
@@ -1,0 +1,193 @@
+import React from 'react';
+
+import classNames from 'classnames';
+
+import { Theme, Themes } from 'LumX/components';
+
+import { COMPONENT_PREFIX } from 'LumX/core/react/constants';
+
+import { ThumbnailVariants } from 'LumX';
+import { Thumbnail } from 'LumX/components/thumbnail/react/Thumbnail';
+import { IGenericProps, getRootClassName } from 'LumX/core/react/utils';
+import { handleBasicClasses } from 'LumX/core/utils';
+
+/**
+ * Authorized size values.
+ */
+enum Sizes {
+    s = 's',
+    m = 'm',
+    l = 'l',
+}
+type Size = Sizes;
+
+/**
+ * Authorized Orientations.
+ */
+enum Orientations {
+    horizontal = 'horizontal',
+    vertical = 'vertical',
+}
+type Orientation = Orientations;
+
+/////////////////////////////
+
+/**
+ * Defines the props of the component.
+ */
+interface IUserBlockProps extends IGenericProps {
+    /* Avatar image. */
+    avatar?: string;
+    /* Simple Action block. */
+    simpleAction?: React.ReactNode;
+    /* Multiple Actions block. */
+    multipleActions?: React.ReactNode;
+    /* Additionnal fields used to describe the use. */
+    fields?: string[];
+    /* User name. */
+    name?: string;
+    /* Orientation. */
+    orientation?: Orientation;
+    /* Size. */
+    size?: Size;
+    /* Theme. */
+    theme?: Theme;
+    /* Callback for the click event. */
+    onClick?(): void;
+    /* Callback for the mouseEnter event. */
+    onMouseEnter?(): void;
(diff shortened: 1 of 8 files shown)


src/components/user-block/react/UserBlock.tsx
```diff
@@ -0,0 +1,193 @@
...
+    s = 's',
+    m = 'm',
+    l = 'l',
+}
+type Size = Sizes;
+
+/**
+ * Authorized Orientations.
+ */
+enum Orientations {
```
mcastagnetti: This enum can be exported in a more global constants since `ImageBlock` will be using it

src/components/user-block/react/UserBlock.tsx
```diff
@@ -0,0 +1,193 @@
...
+    theme: Themes.light,
+};
+/////////////////////////////
+
+/**
+ * Render a user information as a card if orientation is vertical or no action user info block if horizontal.
+ *
+ * @return {React.ReactElement} The component.
+ */
+const UserBlock: React.FC<UserBlockProps> = ({
```
mcastagnetti: `IUserBlockProps` instead of `UserBlockProps` ?

src/components/user-block/react/UserBlock.tsx
```diff
@@ -0,0 +1,193 @@
...
+    className = '',
+    simpleAction,
+    multipleActions,
+}: UserBlockProps): React.ReactElement => {
+    // Special case - When using vertical orientation force the size to be Sizes.l.
+    if (orientation === Orientations.vertical) {
+        size = Sizes.l;
+    }
+
+    const showActions: boolean = orientation === Orientations.vertical;
```
mcastagnetti: `shouldDisplayActions` or something more verbose should be used

src/components/user-block/react/UserBlock.tsx
```diff
@@ -0,0 +1,193 @@
...
+    onClick,
+    onMouseEnter,
+    onMouseLeave,
+    className = '',
+    simpleAction,
+    multipleActions,
+}: UserBlockProps): React.ReactElement => {
+    // Special case - When using vertical orientation force the size to be Sizes.l.
+    if (orientation === Orientations.vertical) {
+        size = Sizes.l;
```
mcastagnetti: Props should not be mutated, maybe we can use a buffer, or force the application of `Sizes.l` whenever we use it

src/components/user-block/react/UserBlock.tsx
```diff
@@ -0,0 +1,193 @@
...
+            className={`${CLASSNAME}__name`}
+            onMouseEnter={onMouseEnter}
+            onMouseLeave={onMouseLeave}
+            onClick={onClick}
+            tabIndex={onClick ? 0 : -1}
+        >
+            {name}
+        </span>
+    ) : (
+        undefined
```
mcastagnetti: `null` or directly using the `name && <span>...</span>` ?

mcastagnetti: Also we more commonly use this notation 
```js
// JsDoc
const displayNameBlock () => <span>...</span>

// ...

{name && displayNameBlock()}
```

src/components/user-block/react/UserBlock.tsx
```diff
@@ -0,0 +1,193 @@
...
+    );
+
+    const fieldsBlock: React.ReactNode =
```
mcastagnetti: Same as `displayNameBlock`

src/components/user-block/react/UserBlock.tsx
GuillaumeNachury: Yep, you are right.
I created the `ORIENTATIONS` constant  in `LumX/core/react/constants`

src/components/user-block/react/UserBlock.tsx
GuillaumeNachury: That makes sense, I wanted to be consistant with the existing `switch` / `Button` components where a `type` is created from the interface.
done

src/components/user-block/react/UserBlock.tsx
GuillaumeNachury: Yes.
done

src/components/user-block/react/UserBlock.tsx
GuillaumeNachury: good catch.

src/components/user-block/react/UserBlock.tsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 
 import { Theme, Themes } from 'LumX/components';
 
-import { COMPONENT_PREFIX } from 'LumX/core/react/constants';
+import { COMPONENT_PREFIX, ORIENTATIONS } from 'LumX/core/react/constants';
```
mcastagnetti: Shouldn't it be in the `LumX/components/index.ts` file along with  `import { Theme, Themes } from 'LumX/components';` ?

GuillaumeNachury: ok done, although I think it would make more sense to get all our constants in a `constant.js` file.

mcastagnetti: You think type definitions should be in the same file as the constants ?

src/components/user-block/react/UserBlock.tsx
```diff
@@ -38,7 +38,7 @@ interface IUserBlockProps extends IGenericProps {
     /* User name. */
     name?: string;
     /* Orientation. */
-    orientation?: ORIENTATIONS;
+    orientation?: Orientations;
```
mcastagnetti: `Orientation` type instead of `Orientations` enum ?

GuillaumeNachury: done

matmkian:
👍

PR #47, opened 2019-04-19
docs(link)

FlorentDinet:
Didn't managed to give the last demo-block a dark-theme...


FlorentDinet:
@matmkian can you review my work?

And also I didn't manage to give the link with icon a display:inline-bloc to prevent this curious padding to happen:
Capture d’écran 2020-09-04 à 15 35 53

If you have a magic trick...

PR #64, opened 2019-05-20
feat(text-field): add component and demo

Flo5k5:


src/components/text-field/react/TextField.tsx
```diff
@@ -0,0 +1,180 @@
...
+                        // tslint:disable-next-line: jsx-no-lambda
+                        onBlur={(): void => setHasFocus(false)}
+                        onChange={handleChange}
+                        {...props}
+                    />
+                </div>
+
+                {(isValid || hasError) && (
+                    <Icon
+                        className={'__input-validity"'}
```
mcastagnetti: Seems wrong classname

src/components/text-field/react/TextField.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+    return (
+        <div
+            className={classNames(className, handleBasicClasses({ prefix: CLASSNAME, theme }), {
+                [`${CLASSNAME}--has-error`]: !isValid && hasError,
+                [`${CLASSNAME}--has-icon`]: icon,
+                [`${CLASSNAME}--has-label`]: label,
+                [`${CLASSNAME}--has-placeholder`]: placeholder,
+                [`${CLASSNAME}--has-value`]: hasValue,
+                [`${CLASSNAME}--is-disabled`]: isDisabled,
```
mcastagnetti: Can be handled by `handleBasicClasses` but don't know if its the right approach or not :man_shrugging:

src/components/text-field/react/TextField.tsx
```diff
@@ -0,0 +1,180 @@
...
+                [`${CLASSNAME}--is-focus`]: hasFocus,
+                [`${CLASSNAME}--is-valid`]: isValid,
```
mcastagnetti: idem `handleBasicClasses`

src/components/text-field/react/TextField.tsx
Flo5k5: Nice catch

matmkian:
Tests ok!

PR #66, opened 2019-05-21
docs(search)

FlorentDinet:


demo/angularjs/components/search/doc.md
```diff
@@ -0,0 +1,17 @@
+# Search
```
lionelmoi: I think this component should be named 
```suggestion
# Search field
```
Search is too much of a concept to not be accompanied by another name.

demo/angularjs/components/search/doc.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Search allow users to specify a word or a phrase to find particular relevant pieces of content.**
```
lionelmoi: `particular` and `relevant` are redundant, `pieces of content` is the proper way but `relevant content` would be acceptable too.

demo/angularjs/components/search/doc.md
```diff
@@ -0,0 +1,17 @@
...
+
+<demo-block component="search" partial="search"></demo-block>
+
+## Suggestion
+
+<demo-block component="search" partial="suggestion"></demo-block>
+
+## Growing search
+
+Use growing search when space is constraint or when you don't want a prominent search.
```
lionelmoi: - `...space is *a* constraint`
- `don't`use contractions ಠ_ಠ
- `and search does not have to be proeminent` to avoid the `you`

demo/angularjs/components/search-field/doc.md
```diff
@@ -0,0 +1,17 @@
+# Search field
+
+**Search field allow users to specify a word or a phrase to find relevant content.**
```
lionelmoi: FIrst remark, `Search field` must be plural. 
I would stress the nature of the component, it is an input composed of a text field and a button.

demo/angularjs/components/search-field/doc.md
```diff
@@ -0,0 +1,17 @@
...
+
+<demo-block component="search-field" partial="search"></demo-block>
+
+## Suggestion
+
+<demo-block component="search-field" partial="suggestion"></demo-block>
+
+## Growing search field
+
+Use growing search field when space is a constraint or search does not have to be prominent.
```
lionelmoi: Since it is a reference to the title, use Uppercase for `growing search field`

demo/angularjs/components/search-field/doc.md
FlorentDinet: I agree for the nature description but it is more like a simple textfield in most cases and a textfield with a suggestion popover for the global search. So I tend to avoid describing its nature.

PR #68, opened 2019-05-22
 feat(chip): polish chip to new standards

mcastagnetti:
Apply lastest rulsets on chip component

### DIFF
--- a/src/components/chip/react/Chip.tsx
+++ b/src/components/chip/react/Chip.tsx
@@ -3,7 +3,6 @@ import React, { SyntheticEvent } from 'react';
 import classNames from 'classnames';
 
 import isFunction from 'lodash/isFunction';
-import noop from 'lodash/noop';
 
 import { Theme, Themes } from 'LumX/components';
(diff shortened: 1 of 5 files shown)


src/components/chip/react/Chip.tsx
```diff
@@ -155,52 +152,51 @@ const Chip: React.FC<IChipProps> = ({
...
                     theme,
-                    unselected: Boolean(!isActive),
+                    unselected: Boolean(!isSelected),
                 }),
             )}
             role="button"
             tabIndex={isDisabled || !hasOnClick ? -1 : 0}
             onClick={onClick || noop}
-            onKeyPress={onClick}
+            onKeyDown={onEnterPressed(onClick || noop)}
```
arnaudvalle: Maybe something like `onClick ? onEnterPress(onClick) : null` I think when a event prop is null it doesn't even attach the event. With noop it would still attach one but do nothing. What do you think?

mcastagnetti: Right :+1:

PR #69, opened 2019-05-23
feat(chip): add color support

matmkian:


matmkian:
To do in react version

src/components/chip/angularjs/chip_directive.js
```diff
@@ -121,6 +122,24 @@ function ChipDirective() {
         if (transclude.isSlotFilled('label')) {
             ctrl.hasLabel = true;
         }
+
+        const defaultProps = {
+            color: 'dark',
```
arnaudvalle: Just to confirm, `dark` is a color or a theme? Looking at the template, it looks like you wanted to make the default theme `dark`?

src/components/chip/react/__snapshots__/Chip.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`<Chip /> Snapshot should render correctly Chip component 1`] = `
 <a
   LabelComponent="Hello World!"
-  className="lumx-chip lumx-chip--size-m lumx-chip--theme-light lumx-chip--is-unselected"
+  className="lumx-chip lumx-chip--color-dark lumx-chip--size-m lumx-chip--is-unselected"
```
arnaudvalle: So no theme by default is actually what we want?

src/components/chip/angularjs/chip_directive.js
mcastagnetti: Don't know about angular version, @matmkian ?

src/components/chip/react/__snapshots__/Chip.test.tsx.snap
mcastagnetti: Yep :3

src/components/chip/angularjs/chip_directive.js
matmkian: No more default value for theme prop
Add color prop with dark default value -> dark is a theme but also a color :-)

arnaudvalle: :wtfception:

matmkian:
Test ok!

PR #70, opened 2019-05-23
feat(thumbnail): add align parameter support

matmkian:


matmkian:
To do in react version
Breaking change -> caption align renamed to align on image block component

PR #71, opened 2019-05-23
Added item selection behavior (via click or navigation) + minor fixes on the list

GuillaumeNachury:


src/components/list-item/react/ListItem.tsx
```diff
@@ -122,6 +150,9 @@ const ListItem: React.FC<ListItemProps> = ({
                 handleBasicClasses({ prefix: CLASSNAME, theme, selected: isSelected, clickable: isClickable, size }),
             )}
             tabIndex={isClickable ? 0 : -1}
+            onFocusCapture={preventParentFocus}
+            onClick={onItemSelected}
+            onKeyDown={onKeyDown()}
```
arnaudvalle: parens really wanted here?

GuillaumeNachury: Yes because `onKeyDown()` actually returns a function that will be used as a handler only it the  a callback has been defined. ~ function composition / curried function approach.

matmkian:
Test ok !

PR #75, opened 2019-05-24
Added arrow on tooltip

GuillaumeNachury:
When using the `popover` tooltip mode, an arrow is now pointing toward the reference element.


FlorentDinet:
I do not see arrows on tooltip mode:
Capture d’écran 2019-05-27 à 10 25 38

FlorentDinet:
Seems ok to me.
The only warning I've got is for the default tooltip style, it's should be:
const demoPopperStyle: CSSProperties = {
    fontSize: '12px',
};
in popover/tooltip.tsx

GuillaumeNachury:
ok I change the demo file

PR #84, opened 2019-06-21
fix(image-block): remove useless export

Flo5k5:


arnaudvalle:
Are you sure it's unused? doing a quick search I can see it in https://github.com/lumapps/design-system/blob/b04e4d75f9fedceae442b4cdca371add12b2e3bf/demo/react/components/lightbox/default.tsx ?

Flo5k5:
> Are you sure it's unused? doing a quick search I can see it in https://github.com/lumapps/design-system/blob/b04e4d75f9fedceae442b4cdca371add12b2e3bf/demo/react/components/lightbox/default.tsx ?

Yep it's not exported anymore since Marco's refactoring. It's now alignment. I will propagate this modification. Thanks for pointing me this issue.

PR #86, opened 2019-06-21
Create dedicated core package and add icon publishing

pidupuis:


pidupuis:
`webpack.umd.js` was the term already in use, and I made a guess that it was a choice, and a collective one, but I totally agree that this is confusing. I couldn't understand it at first read so I'm okay to rename it.

About the duplication, this is basically the main part of my modification. Angularjs and react configurations were pretty efficiently fatorised, but in the end there won't be angularjs anymore. So I voluntarily dissociated the configuration (consequently duplicated the code) to be able to remove angularjs later but to keep the DS isofunctional until then.

arnaudvalle:
@pidupuis ok I see your point. Hopefully AngularJS will be dropped soon enough otherwise any change to the toolchain will have to be done multiple times and you can bet it will be forgotten in one place or another most of the time.

arnaudvalle:
> `webpack.umd.js` was the term already in use, and I made a guess that it was a choice, and a collective one, but I totally agree that this is confusing. I couldn't understand it at first read so I'm okay to rename it.

If you don't get it and you're one of the maintainers then I'm not sure how other contributors will get it? :) I'd say change it (assuming it's trivial - otherwise just do it later in a separate PR it's not a biggie).

pidupuis:
I changed the name to `webpack.prod.js`

PR #106, opened 2019-07-17
fix(expansion-panel): fix default theme

gcornut:
- Add default theme prop
- Better handle theme prop to ignore invalid theme value


src/components/expansion-panel/react/ExpansionPanel.tsx
```diff
@@ -115,15 +116,14 @@ const ExpansionPanel: React.FC<ExpansionPanelProps> = (props: ExpansionPanelProp
         setShowWrapper(shouldOpen);
     };
 
-    const color = theme === Theme.light ? ColorPalette.dark : ColorPalette.light;
+    const color = theme === Theme.dark ? ColorPalette.light : ColorPalette.dark;
```
Flo5k5: 😂

gcornut: Yeah if the theme is dark, the font is light and otherwise the font is dark

PR #113, opened 2019-07-23
feat(grid): add angularjs component

matmkian:


matmkian:
Need change classes on React size + change gutter prop type from integer to string (regular/big/huge)

PR #116, opened 2019-07-25
feat(icon): add color variant parameter

matmkian:
Add color variant parameter

To do in React version :
Add color-variant prop (no default value = if no prop, no class)
Associated class: lumx-icon--color-variant-{colorVariant} (N, L2, L5...)


matmkian:
test ok, need to resolve conflict

PR #118, opened 2019-08-01
feat(react-popover): rework popover component to remove popperjs

Flo5k5:


src/components/popover/react/Popover.tsx
```diff
@@ -119,216 +106,291 @@ const CLASSNAME: string = getRootClassName(COMPONENT_NAME);
...
-                    }}
-                </Popper>
+    /**
+     * Handle mouse leaving the popover.
+     */
+    const handleMouseLeave = (): void => {
+        setIsMouseEntered(false);
+    };
+
+    useEffect(() => {
```
hchevallier: Why not put all this useEffect in the useComputePosition ?
If you do so, you can get rid of the "value" fake state

gcornut: Also, we could try to externalize the state to make the component fully controlled.
In the demo we could initialize the Popover with a position (x, y, width, height) instead of the anchorRef and call the useCompustePosition externally.

demo/react/components/popover/controled.tsx
```diff
@@ -32,47 +35,50 @@ const demoPopoverHolderStyle: CSSProperties = {
...
     const toggleTooltipDisplay = (): void => {
         setTooltipDisplayed(!isTooltipDisplayed);
     };
 
+    const { computedPosition, isVisible } = Popover.useComputePosition(
+        Placement.RIGHT_END,
+        anchorRef,
+        popoverRef,
+        isTooltipDisplayed,
+    );
```
gcornut: Nice extraction/separation of logic here :)

src/components/user-block/react/UserBlock.tsx
```diff
@@ -83,84 +83,84 @@ const DEFAULT_PROPS: IDefaultPropsType = {
...
-                </div>
-            )}
-            {shouldDisplayActions && simpleAction && <div className={`${CLASSNAME}__action`}>{simpleAction}</div>}
-            {shouldDisplayActions && multipleActions && (
-                <div className={`${CLASSNAME}__actions`}>{multipleActions}</div>
-            )}
-        </div>
-    );
-};
+const UserBlock = React.forwardRef(
```
gcornut: Is there a reason for using `forwardRef` over creating a new `userBlockRef` prop like it is done in other components?

src/components/popover/react/Popover.tsx
```diff
@@ -40,54 +34,44 @@ enum PopperPlacement {
...
-interface IPopperOffsets {
+/**
+ * Vertical and horizontal offset of the popover.
+ */
+interface IOffsets {
     vertical?: number;
     horizontal?: number;
 }
-type PopperOffsets = IPopperOffsets;
+type Offsets = IOffsets;
```
gcornut: The plural is not necessary here I think. `IOffset` and `Offset` would be better I think

src/components/user-block/react/UserBlock.tsx
hchevallier: No problem
I replaced it by a userBlockRef prop

src/components/popover/react/Popover.tsx
hchevallier: Done

PR #119, opened 2019-08-02
feat(dropdown): react component

Flo5k5:
Need Popover to be merged before.

Missing test, check if the clickaway callback is triggerred when closeonclick is set to true. The problem is the eventlistener is on the document while in the jest context the mounted component is not attached to document so it won't bubble. 
I had some troubles to test this case and it already took me too much time so if anyone has an idea, I'll add the test.

### DIFF
--- a/src/components/button/react/ButtonGroup.tsx
+++ b/src/components/button/react/ButtonGroup.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, ReactNode, Ref, forwardRef } from 'react';
+import React, { ReactElement, ReactNode, Ref } from 'react';
 
 import classNames from 'classnames';
 
@@ -12,7 +12,10 @@ import { Button } from './Button';
 /**
  * Defines the props of the component
  */
-interface IProps extends IGenericProps {}
+interface IProps extends IGenericProps {
+    /* Ref passed to the wrapper. */
+    buttonGroupRef?: Ref<HTMLDivElement>;
+}
 type ButtonGroupProps = IProps;
 
 /////////////////////////////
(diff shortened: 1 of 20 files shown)


src/components/dropdown/react/Dropdown.tsx
```diff
@@ -75,73 +76,68 @@ const DEFAULT_PROPS: IDefaultPropsType = {
...
+        } else {
+            window.removeEventListener('keydown', onEscapePressed(onClose!));
         }
-
-        setIsOpen(showDropdown);
-    }, [showDropdown]);
+        return (): void => {
+            window.removeEventListener('keydown', onEscapePressed(onClose!));
+        };
+    }, [showDropdown, escapeClose]);
```
gcornut: The `onClose` is also a dependency of this `useEffect` and it might be `undefined` or not a function. Perhaps we could check it's value here before defining the `useEffect`?

src/components/dropdown/react/Dropdown.tsx
```diff
@@ -62,6 +62,7 @@ const CLASSNAME: string = getRootClassName(COMPONENT_NAME);
 const DEFAULT_PROPS: IDefaultPropsType = {
     closeOnClick: true,
     escapeClose: true,
```
gcornut: Rename to `closeOnEscape` to uniformize with `closeOnClick` ?

src/components/button/react/ButtonGroup.tsx
```diff
@@ -74,19 +74,17 @@ function _validate(props: ButtonGroupProps): ReactNode {
...
-}: ButtonGroupProps): ReactElement => {
-    const newChildren: ReactNode = _validate({ children });
-
-    return (
-        <div className={classNames(className, CLASSNAME)} {...props}>
-            {newChildren}
-        </div>
-    );
-};
+const ButtonGroup: React.FC<ButtonGroupProps> = forwardRef(
```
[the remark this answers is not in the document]
hchevallier: Done

src/components/dropdown/react/Dropdown.tsx
hchevallier: You're right

src/components/dropdown/react/Dropdown.tsx
hchevallier: Good point !

PR #127, opened 2019-08-20
fix(react-expansion-panel): minor fixes

gcornut:
- Fix max height not adapting to the children height changing
- Convert to fully controlled component
- Add missing props forwarding n `DragHandle` component (required for react-beautiful-dnd

### DIFF
--- a/demo/react/components/expansion-panel/label.tsx
+++ b/demo/react/components/expansion-panel/label.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState } from 'react';
 
 import { ExpansionPanel, Theme } from 'LumX';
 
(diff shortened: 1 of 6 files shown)


demo/react/components/expansion-panel/label.tsx
```diff
@@ -13,14 +13,18 @@ interface IProps {
...
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis eu sem et mauris convallis tempor. Mauris
-            placerat enim eget ligula fermentum, in aliquam lorem congue. Vivamus lacinia consectetur mollis.
-        </ExpansionPanel>
-    </>
-);
+const DemoComponent: React.FC<IProps> = ({ theme }: IProps): ReactElement => {
+    const [isOpen, setIsOpen] = useState(false);
+
+    return (
+        <>
```
hchevallier: You can remove the React.Fragment (<>) here

gcornut:
I'm also adding props forwarding for the `DragHandle` component in this fix. It's quite critical to correctly use the react-beautiful-dnd with it

PR #139, opened 2019-09-05
feat(demo-react): manual prop table section (instead of the automatic one)

gcornut:
This PR removes the automatic insertion of the component properties tables in favor of manually inserting the prop table on the demo.


FlorentDinet:
Everything fine for me 👌

PR #150, opened 2019-09-06
docs(table): update from angular

FlorentDinet:
demo block note working (might be template string thing again)


gcornut:
We can't make the template strings work correctly in MDX. We are looking into it with @pidupuis but for now, please use simple string concatenation (ex: `key={'tr-' + key}`)

Also, it might be because you forgot to commit it but `orderBy` is not imported, `setTable` and `dataTableBody` are not defined.

I guess you wanted to do something like:
```
import { useState } from 'react';
....
import { orderBy } from 'lodash'

....
const [dataTableBody, setTable] = useState(tableBody);
....
```

FlorentDinet:
I really tried, but I must admit that's way beyond my skills. If a real dev. can take from here, It'll be better for everyone's sake :D

gcornut:
Ok I can do it

gcornut:
It was indeed not so simple. 
@FlorentDinet  I let you review/test that the demo works as expected

PR #153, opened 2019-09-10
refactor(RadioButton): simplify props

gcornut:
Changelog radio button:
- Remove dedicated RadioButtonLabel and RadioButtonHelper component (use prop instead)
- Change onChange handler to return the valu directly without a wrapping object
- Remove old tsx demo
- Update mdx demo to be interactive


pidupuis:
Test OK

PR #160, opened 2019-09-17
fix(dropdown): fix onClose unnecessary calls

hchevallier:
**Bug**
Fix https://lumapps.atlassian.net/browse/LUM-5350

**Fix**
- Add a condition to prevent clickAway from calling onClose when the dropdown is already closed.
- Fix the eventListener removal on the Escape press handler.

### DIFF
--- a/src/components/dropdown/react/Dropdown.tsx
+++ b/src/components/dropdown/react/Dropdown.tsx
@@ -125,21 +125,19 @@ const Dropdown: React.FC<DropdownProps> = ({
         </div>
     );
 
-    const onEscapeHandler = onClose ? onEscapePressed(onClose) : undefined;
+    const onEscapeHandler = closeOnEscape && onClose && showDropdown && onEscapePressed(onClose);
 
     useEffect(() => {
-        if (!onEscapeHandler) {
+        if (!onEscapeHandler || !wrapperRef.current) {
             return undefined;
         }
         if (closeOnEscape && showDropdown && wrapperRef.current) {
             window.addEventListener('keydown', onEscapeHandler);
-        } else {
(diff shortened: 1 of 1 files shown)


src/components/dropdown/react/Dropdown.tsx
```diff
@@ -125,31 +125,33 @@ const Dropdown: React.FC<DropdownProps> = ({
         </div>
     );
 
+    const onEscapeHandler = onClose ? onEscapePressed(onClose) : undefined;
```
gcornut: Simplifiable into `const onEscapeHandler = closeOnEscape && onClose && onEscapePressed(onClose);` ?

src/components/dropdown/react/Dropdown.tsx
```diff
@@ -125,31 +125,33 @@ const Dropdown: React.FC<DropdownProps> = ({
...
+
     useEffect(() => {
-        if (!onClose) {
+        if (!onEscapeHandler) {
```
gcornut: You can check the `showDropdown` and `wrapperRef.current` variables here

src/components/dropdown/react/Dropdown.tsx
```diff
@@ -125,31 +125,33 @@ const Dropdown: React.FC<DropdownProps> = ({
...
             return undefined;
         }
         if (closeOnEscape && showDropdown && wrapperRef.current) {
-            window.addEventListener('keydown', onEscapePressed(onClose));
+            window.addEventListener('keydown', onEscapeHandler);
         } else {
```
gcornut: The else branch is not necessary. The useEffect will automatically call your `removeEventListener` inside your clean up lambda if the dependencies change

PR #162, opened 2019-09-18
feat(post-block): add the possibility of html injection in the text prop

BanshISquare:
Mirror of how it has been done on the ImageBlock component


matmkian:
Not ok for me -> should be raw text as this is a truncated excerpt

PR #191, opened 2019-10-14
feat(autocomplete simple): react component for autocomplete - simple variation

juanigalan91:
Created the first version of the autocomplete component. it supports:
- Filter options based on a given input
- Control over the value, options, and when the suggestions are displayed or not
- Navigate between options using the keyboard
- Selecting one option either by clicking or pressing Enter

**Simple usage**
QHePgrkFJV

**Keyboard navigation**
BrMiLlz37o

**No results**
BRTzGGX353

**Note**: this variation may need css changes since the `no results` label is a little bit more gray on the mockups.

**Leave and come back**
96ohhIcDx4

Side effects:
- ListItem: in order to allow the keyboard navigation, it was necessary to add some sort of mechanism to ListItem in order to, from the outside, control which ListItem is highlighted. To do that, we can create a new prop called `isHighlighted` and control the boolean from the outside
- TextField: added textFieldRef in order to forward the ref, since `Dropdown` needs that reference to display the popover. Also, TextField did not allow to override `onFocus` and `onBlur`, so now, when the `onFocus` and `onBlur` are executed inside the component, they will check whether there are functions passed in as props and they will be executed first. `inputRef` was also added since it is necessary to have a reference to the actual input, and not the wrapper as `textFieldRef` did.

### DIFF
--- a/demo/react/doc/product/components/autocomplete.mdx
+++ b/demo/react/doc/product/components/autocomplete.mdx
@@ -47,6 +47,8 @@ import { DOWN_KEY_CODE, ENTER_KEY_CODE, TAB_KEY_CODE, UP_KEY_CODE } from 'LumX/c
 
     const [showSuggestions, setShowSuggestions] = React.useState(false);
     const [filterValue, setFilterValue] = React.useState('');
+    const inputRef = React.useRef(null);
+
     const filteredCities = CITIES.filter(city => city.text.replace(' ', '').toLowerCase().includes(filterValue.replace(' ', '').toLowerCase()));
     const closeAutocomplete = () => setShowSuggestions(false);
 
@@ -62,17 +64,14 @@ import { DOWN_KEY_CODE, ENTER_KEY_CODE, TAB_KEY_CODE, UP_KEY_CODE } from 'LumX/c
 
     const {
         activeItemIndex,
-        onKeyboardNavigation,
-        setActiveItemIndex,
-        resetActiveIndex
     } = List.useKeyboardListNavigation(
         filteredCities,
         setSelectedCity,
+        inputRef,
     );
 
     const onFocus = (evt) => {
-        resetActiveIndex();
-        return setShowSuggestions(filterValue.length > 0);
+        setShowSuggestions(filterValue.length > 0);
     }
 
     return (
@@ -82,8 +81,8 @@ import { DOWN_KEY_CODE, ENTER_KEY_CODE, TAB_KEY_CODE, UP_KEY_CODE } from 'LumX/c
             onClose={closeAutocomplete}
             value={filterValue}
             onChange={onChange}
-            onKeyDown={onKeyboardNavigation}
             onFocus={onFocus}
+            inputRef={inputRef}
         >
             <List isClickable>
                 { filteredCities.length > 0 ? filteredCities.map((city, index) => (
--- a/demo/react/doc/product/components/autocomplete.mdx
+++ b/demo/react/doc/product/components/autocomplete.mdx
@@ -89,16 +89,19 @@ import { DOWN_KEY_CODE, ENTER_KEY_CODE, TAB_KEY_CODE, UP_KEY_CODE } from 'LumX/c
         }
     };
 
+    const onFocus = (evt) => setShowSuggestions(filterValue.length > 0);
+    const onBlur = (evt) => setShowSuggestions(false);
+
     return (
         <Autocomplete
             theme={theme}
-            showSuggestions={showSuggestions}
-            closeOnClick
-            closeOnEscape
+            isOpen={showSuggestions}
             onClose={closeAutocomplete}
             value={filterValue}
             onChange={onChange}
             onKeyDown={onKeyDown}
+            onFocus={onFocus}
+            onBlur={onBlur}
         >
             <List isClickable>
                 { filteredCities.length > 0 ? filteredCities.map((city, index) => (
--- a/demo/react/doc/product/components/autocomplete.mdx
+++ b/demo/react/doc/product/components/autocomplete.mdx
@@ -1,5 +1,6 @@
 ```javascript import
-import { Autocomplete } from 'LumX';
+import { Autocomplete, List, ListItem, ListItemSize, Size, TextField, Chip, ListDivider, Icon, ListSubheader } from 'LumX';
+import { useBooleanState } from 'LumX/core/react/hooks';
 ```
 
 # Autocomplete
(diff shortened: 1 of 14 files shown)


src/components/list/react/ListItem.tsx
```diff
@@ -147,6 +152,7 @@ const ListItem: React.FC<ListItemProps> = ({
             onFocusCapture={preventParentFocus}
             onClick={onItemSelected}
             onKeyDown={onKeyDown()}
+            data-focus-visible-added={isHighlighted || undefined}
```
juanigalan91: I am not 100% sure of this change, seems a little bit hacky to use the `data-focus-visible-added` to manage the highlight. Maybe we can create another css class that has those css styles that the selector `[data-focus-visible-added]` has and then use the `isHighlighted` prop to set that class. What do you guys think?

src/components/autocomplete/react/Autocomplete.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Whether the suggestions from the autocomplete should be displayed or not.
+     * Useful to control when the suggestions are displayed from outside the component
+     */
+    showSuggestions: boolean;
+
+    /**
+     * Whether a click anywhere out of the Autocomplete would close it
+     * @see {@link TextFieldProps#closeOnClick}
```
gcornut: I think you meant `DropdownProps#closeOnClick`

src/components/autocomplete/react/Autocomplete.tsx
```diff
@@ -0,0 +1,138 @@
...
+     */
+    closeOnClick?: boolean;
+
+    /**
+     * Whether an escape key press would close the Autocomplete.
+     * @see {@link TextFieldProps#closeOnEscape}
```
gcornut: I think you meant DropdownProps#closeOnEscape

src/components/autocomplete/react/Autocomplete.tsx
```diff
@@ -0,0 +1,138 @@
...
+     * Text field value.
+     * @see {@link TextFieldProps#onChange}
+     */
+    value: string;
+
+    /**
+     * Whether the suggestions from the autocomplete should be displayed or not.
+     * Useful to control when the suggestions are displayed from outside the component
+     */
+    showSuggestions: boolean;
```
gcornut: Maybe `isOpen` would be simpler here and pairs well with the `onClose` handler.

src/components/autocomplete/react/Autocomplete.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { handleBasicClasses } from 'LumX/core/utils';
+import { IGenericProps, getRootClassName } from 'LumX/react/utils';
+
+/////////////////////////////
+
+/**
+ * Defines the props of the component.
+ */
+interface IAutocompleteProps extends IGenericProps {
+    /** Children of the Autocomplete. */
```
gcornut: Maybe you could describe the children as being the autocomplete suggestions

src/components/autocomplete/react/Autocomplete.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * The <div> element that holds the dropdown content.
+     */
+    dropdown: Wrapper;
+
+    props: ISetupProps;
+
+    /**
+     * [Enter the description of this wrapper].
+     * [You should also probably change the name of the wrapper to something more meaningful].
```
gcornut: To update

src/components/autocomplete/react/Autocomplete.tsx
```diff
@@ -0,0 +1,138 @@
...
+                textFieldRef={textfieldRef}
+                onKeyDown={onKeyDown}
+                {...forwardedProps}
+            />
+            <Dropdown
+                anchorRef={textfieldRef}
+                showDropdown={showSuggestions}
+                closeOnClick={closeOnClick}
+                closeOnEscape={closeOnEscape}
+                {...forwardedProps}
```
gcornut: We forward props to one element at a time. Generally its the main element wrapped by a component (a native DOM element or wrapped component).
For the autocomplete, we have two wrapped component, I would say the forwarded props should be on the div.
You do have to forward more props (like the `onClose`) but that way it's all listed in the props interface documentation.

src/components/list/react/ListItem.tsx
gcornut: Have you tried using `isActive`? I think it would work for you
The List component might override the `isActive` props on ListItem but it should be fixable

src/components/text-field/react/TextField.tsx
```diff
@@ -48,6 +48,9 @@ interface ITextFieldProps extends IGenericProps {
     /** Theme. */
     theme?: Theme;
 
+    /** A ref that will be passed to the wrapper element. */
+    textFieldRef?: Ref<HTMLDivElement>;
```
gcornut: :+1:

src/components/autocomplete/react/Autocomplete.tsx
```diff
@@ -0,0 +1,138 @@
...
+const DEFAULT_PROPS: Partial<AutocompleteProps> = {
+    closeOnClick: true,
+    closeOnEscape: true,
+    showSuggestions: undefined,
+};
+
+/////////////////////////////
+
+/**
+ * [Enter the description of the component here].
```
gcornut: To update

demo/react/doc/product/components/autocomplete.mdx
```diff
@@ -0,0 +1,130 @@
...
+            id: 'lyon'
+        },
+        {
+            text: 'Montevideo',
+            id: 'montevideo'
+        }
+    ];
+
+    const [showSuggestions, setShowSuggestions] = React.useState(false);
+    const [activeIndex, setActiveIndex] = React.useState(-1);
```
gcornut: I think it could be possible to handle the active element index in a custom hook that we could use both here and in the List component.

We might have to refactor the List/ListItem components to extract there logic

We could do something like this: 
```jsx
    // Listen to keydown on textField to handle keyboard selection.
    const [activeIndex, onSelectedItem] = useKeyboardSelection(textFieldRef);
    onSelectedItem(selectedIndex => /* do something */); 

    return (
        <Autocomplete ... >
            <List isClickable>
                {filteredCities.map((city, index) => (
                        <ListItem isActive={index === activeIndex}>
                            <div>{city.text}</div>
                        </ListItem>
```

demo/react/doc/product/components/autocomplete.mdx
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    };
+
+    return (
+        <Autocomplete
+            theme={theme}
+            showSuggestions={showSuggestions}
+            closeOnClick
+            closeOnEscape
```
gcornut: These are true by default in the dropdown, I don't think you need them here

src/components/list/react/ListItem.tsx
gcornut: We could rename to `isHover` using a class name `list-item--hovered`

demo/react/doc/product/components/autocomplete.mdx
juanigalan91: We decided to go for the custom hook but for now use it on the Autocomplete component, in order to have a lower impact on the product. later on, it would be great to have that hook implemented on the List as well.

juanigalan91: @gcornut I uploaded the first version of the custom hook!

demo/react/doc/product/components/autocomplete.mdx
```diff
@@ -0,0 +1,114 @@
+```javascript import
+import { Autocomplete, List, ListItem, ListItemSize, Size, TextField, Chip, ListDivider, Icon, ListSubheader } from 'LumX';
+import { useBooleanState } from 'LumX/core/react/hooks';
+import { DOWN_KEY_CODE, ENTER_KEY_CODE, TAB_KEY_CODE, UP_KEY_CODE } from 'LumX/core/constants';
```
gcornut: There is many unused imports here

juanigalan91: yeah, I removed them on the next component: https://github.com/lumapps/design-system/pull/194/commits/983a0a7616cfa14e2de9caf1ec8150028a8177d6

matmkian:
I don't understand why we don't focus the list and focus elements on keyboard navigation -> check what have been done in angularjs on dropdown.

juanigalan91:
We can discuss it in our meeting today!

juanigalan91:
Just for visibility purposes, we discussed and the reason why we need the `isHighlighted` prop is that if we move the focus to the list, we loose it on the text field, so the user cannot continue typing. By taking this approach, we can "highlight" a result while maintaining focus on the text field

FlorentDinet:
Nice work !
My only doubt come from the "no result" which is indeed to close to an actual "suggestion item"
Capture d’écran 2019-10-28 à 14 57 38
I'll suggest to display nothing, when there is no suggestion

juanigalan91:
Awesome, I can definitely do that!

FlorentDinet:
Nice
I did catch something when hitting backspace key :
Oct-28-2019 16-27-47

losing focus on backspace

juanigalan91:
@FlorentDinet fixed ! thanks for the catch !

FlorentDinet:
ok, last thing I saw is that we're kind of "stuck" in the field with keyboard navigation:
- can't hit "tab" to focus next input
- can't hit "shift+tab" to focus previous input

I'd understand if tab was used to navigate between suggestions but that's not the case

Anyway it's not blocking, just icing on the cake

FlorentDinet:
👌

PR #194, opened 2019-10-16
feat(autocomplete-multiple): React component and tests

juanigalan91:
**Note**: this PR heavily depends on PR #191 , in order to merge it, we need to finish review and merge that one first.

Created a multiple autocomplete component. It provides the following functionality:
- Select multiple chips from a list of values
- Allows new keyboard events such as backspace in order to remove the selected chips
- Keeps all the functionality coming from the Simple Autocomplete component.

Use Cases:

**Multiple Selection with keyboard and mouse navigation**
znbEL8INfv

**Creating new values**
QQ5buyRZD1

**Deleting selected values**
FqFMH1hR4G

**Keyboard navigation**
lmUI1rkjhe

Side effects from the implementation:
- `Chip` now has a `isHighlighted` prop that allows to set the css class into a focused state in order to allow deletion from the Autocomplete
- `TextField` now receives a prop called `before` (similar to what Chip has) that allows to add a component before the input. With this, we can add the different selected chips before the input.
- A refactor was done on the `useKeyboardListNavigation` hook in order to allow a simpler maintenance in the future, with regards to events tracked while using the hook

**Note**: there might be some css changes needed in order to complete the component:
- Adjust spaces between chips
- Adjust dropdown width in order to correspond to Zeplin specification
- Make chips go down the text field and not to the sides when there are a lot of chips selected

### DIFF
--- a/src/components/chip/react/Chip.tsx
+++ b/src/components/chip/react/Chip.tsx
@@ -30,6 +30,8 @@ interface IChipProps extends IGenericProps {
     /* Whether the chip has pointer on hover. */
     isClickable?: boolean;
     /** Indicates if the chip is currently in an active state or not. */
+    isHighlighted?: boolean;
+    /** Indicates if the chip is currently in a highlighted state or not. */
     isSelected?: boolean;
     /** Indicates if the chip is currently disabled or not. */
     isDisabled?: boolean;
@@ -78,6 +80,7 @@ const DEFAULT_PROPS: IDefaultPropsType = {
     color: ColorPalette.dark,
     isClickable: false,
     isDisabled: false,
+    isHighlighted: false,
     isSelected: false,
     size: Size.m,
     theme: undefined,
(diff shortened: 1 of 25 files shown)


src/components/chip/react/Chip.tsx
```diff
@@ -29,7 +29,9 @@ interface IChipProps extends IGenericProps {
     color?: Color;
     /* Whether the chip has pointer on hover. */
     isClickable?: boolean;
-    /** Indicates if the chip is currently in an active state or not. */
+    /** Indicates if the chip is currently in a highlighted state or not. */
+    isHighlighted?: boolean;
```
matmkian: Pourquoi ajouter ça ? Le simple fait de focus la chip ne suffit pas ?

juanigalan91: non, parce qu'on perd le focus sur le text field si on fait de focus en la chip, donc l'utilisateur ne peut pas continuer à écrire dans le texte field.

juanigalan91:
Closing this one since there were several changes after this implementation.

PR #196, opened 2019-10-21
feat(dialog): add is loading prop support

matmkian:
Need React part


matmkian:
Test ko missing class `lumx-dialog--is-loading` when is loading

PR #199, opened 2019-10-21
feat(textfield-enhancements): Text field enhancements

juanigalan91:
Due to the upcoming feature `Autocomplete and Autocomplete multiple` there are some changes that we need on the Text Field:
- In order to have the possibility to show a list of chips before the Text Field, we add the props `chips` to the Text Field and show them before the native input.

Screenshot 2019-10-25 at 10 32 56

- In order to have the ability to listen to the `focus` and `blur` events on the text field itself, it was necessary to add the props `onFocus` and `onBlur` and pass them on the native input, where they were overwritten. If these functions are passed on, they will be executed.

- Furthermore, in order to use this component in conjunction  with the Dropdown, we need to provide a `ref` to the Dropdown so it knows where to open. On the other hand, if we want to manually trigger the focus on the text field, we need a reference to the input itself (input or text area, depends on the type of Text Field used). In order to tackle that, two different props are created, one of them associated with the Text field wrapper, and another one associated with to the native input. 

- Also, a prop called `isClearable` was added in order to allow erasing the content on the text field.
ECmxFl0MH2

- Moreover, a component called ChipGroup was created in order to support a list of chips presented in a grouped fashion, which we will use in the future for the `AutocompleteMultiple` component

### DIFF
--- a/CHANGELOG.react.md
+++ b/CHANGELOG.react.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Fix stroke-width property on react progress bar.
 -   Fixes to Text Area related to Input resizing and `valid` and `error` states in Material
+-   Text field enhancements (`before`, `after`, `onFocus`, `onBlur`, `textfieldRef`, `inputRef` props) [#199](https://github.com/lumapps/design-system/pull/199)
 
 ## [0.11.0][] - 2019-10-23
 
--- a/CHANGELOG.react.md
+++ b/CHANGELOG.react.md
@@ -7,11 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+-   Text field enhancements (`chips`, `onFocus`, `onBlur`, `textfieldRef`, `inputRef` props) [#199](https://github.com/lumapps/design-system/pull/199)
+
 ### Changed
 
 -   Fix stroke-width property on react progress bar.
 -   Fixes to Text Area related to Input resizing and `valid` and `error` states in Material
--   Text field enhancements (`before`, `after`, `onFocus`, `onBlur`, `textfieldRef`, `inputRef` props) [#199](https://github.com/lumapps/design-system/pull/199)
 
 ## [0.11.0][] - 2019-10-23
 
--- a/demo/react/doc/product/components/chip.mdx
+++ b/demo/react/doc/product/components/chip.mdx
@@ -1,6 +1,6 @@
 ```javascript import
 import { useState } from 'react';
-import { Chip, Icon, Size } from 'LumX';
+import { Chip, ChipGroup, Icon, Size } from 'LumX';
 import { mdiEmail, mdiClose, mdiCloseCircle, mdiMenuDown, mdiFilterVariant } from 'LumX/icons';
 ```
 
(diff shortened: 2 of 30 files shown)


juanigalan91:
@pidupuis @gcornut we can once again do a review on this one, sorry for the long list of commits !

demo/react/doc/product/components/chip.mdx
```diff
@@ -140,6 +140,32 @@ Use small when space is a constraint, as in areas like text fields, selects, dro
...
+                Dismissible
+            </Chip>
+        </ChipGroup>
+    );
+};
+```
+
+# Properties
+
+<PropTable component="ChipGroup" />
```
gcornut: You can probably add the Chip prop table here too

CHANGELOG.react.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Fix stroke-width property on react progress bar.
 -   Fixes to Text Area related to Input resizing and `valid` and `error` states in Material
+-   Text field enhancements (`before`, `after`, `onFocus`, `onBlur`, `textfieldRef`, `inputRef` props) [#199](https://github.com/lumapps/design-system/pull/199)
```
gcornut: It could be under `Unreleased > Added` rather than `Changed`, what do you think?

src/components/chip/react/ChipGroup.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * The default value of props.
+ */
+const DEFAULT_PROPS: IDefaultPropsType = {
+    align: 'left',
+};
+
+/**
+ * The display name of the component.
+ */
+const COMPONENT_NAME = `${COMPONENT_PREFIX}Chip`;
```
gcornut: s/Chip/ChipGroup/

src/components/chip/react/ChipGroup.tsx
```diff
@@ -0,0 +1,83 @@
...
+    className,
+    align = DEFAULT_PROPS.align,
+    children,
+    ...props
+}: ChipGroupProps): ReactElement => {
+    const chipGroupClassName = classNames(
+        `${CSS_PREFIX}-chip-group`,
+        `${CSS_PREFIX}-chip-group--align-${align}`,
+        className,
+    );
```
gcornut: The component classes can be generated with:
```jsx
handleBasicClasses({
                    align,
                    prefix: CLASSNAME,
                }),
```

CHANGELOG.react.md
gcornut: Also I don't see the `before` and `after` props in `ITextFieldProps`

src/components/text-field/react/TextField.tsx
```diff
@@ -22,6 +22,9 @@ enum TextFieldType {
  * Defines the props of the component.
  */
 interface ITextFieldProps extends IGenericProps {
+    /** A Chip Group to be rendered before the main text input */
+    chips?: HTMLElement | ReactNode;
```
hchevallier: Stop me if I'm wrong but, if the goal is to have a TextField with multiple values, why not pass an array to the `value` prop ?
If `value` is an array, then I render a ChipGroup into the TextField.

juanigalan91: Because they are separate things that will be rendered. On the one hand, we have the list of chips, which are already selected values. On the other hand, we have the value itself, which is the text to be displayed on the Text Field. Imagine the Tagz component, that has a list of chips on the left and a text on the right.

juanigalan91: Screenshot 2019-10-25 at 12 04 32

gcornut: And the `ReactNode` type in typescript is a union type that actually includes `Array<ReactNode>` (yes it's a recursive reference ^^)

hchevallier: Ok :ok_hand:

PR #200, opened 2019-10-22
fix(textarea styles): Text area height and style fixes

juanigalan91:
All around fixes to Text Area component:
- Remove resizing from browsers that [support it](https://caniuse.com/#feat=css-resize) by using `resize` attribute. For the ones that do not support it, we can make the width fixed.
- Allow automatic resizing for the height when the user is writing (with a little bit of magic)
ZPC9shZ3pF

- Add valid and error css to the Text Area component for Material CSS
Screenshot 2019-10-22 at 09 42 45
Screenshot 2019-10-22 at 09 43 25

### DIFF
--- a/src/components/text-field/react/TextField.tsx
+++ b/src/components/text-field/react/TextField.tsx
@@ -53,7 +53,7 @@ interface ITextFieldProps extends IGenericProps {
     useCustomColors?: boolean;
 
     /** Minimum rows to be displayed in a text area. */
-    minimumRows: number;
+    minimumRows?: number;
 
     /** A ref that will be passed to the input or text area element. */
     inputRef?: RefObject<HTMLInputElement> | RefObject<HTMLTextAreaElement>;
--- a/src/components/text-field/react/TextField.tsx
+++ b/src/components/text-field/react/TextField.tsx
@@ -88,10 +88,55 @@ const COMPONENT_NAME = `${COMPONENT_PREFIX}TextField`;
 const CLASSNAME: string = getRootClassName(COMPONENT_NAME);
 
 /**
- * The line height for the Text Area. Please keep in mind that if that changes via css, the prop lineHeight needs to be changed
+ * The minimum number of rows that we want to display on the text area
  */
-const LINE_HEIGHT = 20;
 const MIN_ROWS = 2;
+
+/**
+ * Hook that allows to calculate the number of rows needed for a text area.
+ * @param minimumNumberOfRows Minimum number of rows that we want to display.
+ * @return rows to be used and a callback to recalculate
+ */
+const useComputeNumberOfRows = (
+    minimumNumberOfRows: number,
+): {
+    /** number of rows to be used on the text area */
+    rows: number;
+    /**
+     * Callback in order to recalculate the number of rows due to a change on the text area
+     * @param event Change event.
+     */
+    recomputeNumberOfRows(event: React.ChangeEvent): void;
+} => {
+    const [rows, setRows] = useState(minimumNumberOfRows);
+
+    const recompute = (event: React.ChangeEvent): void => {
+        /**
+         * HEAD's UP! This part is a little bit tricky. The idea here is to only
+         * display the necessary rows on the textarea. In order to dynamically adjust
+         * the height on that field, we need to:
+         * 1. Set the current amount of rows to the minimum. That will make the scroll appear.
+         * 2. With that, we will have the `scrollHeight`, meaning the height of the container adjusted to the current content
+         * 3. With the scroll height, we can figure out how many rows we need to use by dividing the scroll height
+         * by the line height.
+         * 4. With that number, we can readjust the number of rows on the text area. We need to do that here, if we leave that to
+         * the state change through React, there are some scenarios (resize, hitting ENTER or BACKSPACE which add or remove lines)
+         * when we will not see the update and the rows will be resized to the minimum.
+         * 5. In case there is any other update on the component that changes the UI, we need to keep the number of rows
+         * on the state in order to allow React to re-render. Therefore, we save them using `useState`
+         */
+        (event.target as HTMLTextAreaElement).rows = minimumNumberOfRows;
+        const currentRows = event.target.scrollHeight / (event.target.clientHeight / minimumNumberOfRows);
+        (event.target as HTMLTextAreaElement).rows = currentRows;
+
+        setRows(currentRows);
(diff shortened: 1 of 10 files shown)


src/components/text-field/style/lumapps/_index.scss
```diff
@@ -60,7 +60,7 @@
     }
 
     &__input-native {
-        flex: 1;
+        width: 100%;
```
matmkian: Why?

src/components/text-field/style/lumapps/_index.scss
```diff
@@ -84,6 +84,13 @@
         }
 
         textarea {
+            overflow: hidden;
```
matmkian: Did you check on angularjs version?

src/components/text-field/style/material/_index.scss
```diff
@@ -87,3 +87,90 @@
         @include lumx-text-field-material-label-focus('light');
     }
 }
+
+/* Text field validity
+   ========================================================================== */
+
+.#{$lumx-base-prefix}-text-field--theme-light {
```
matmkian: Useless in my mind, we never do validation on textarea.

src/components/text-field/style/material/_mixins.scss
```diff
@@ -56,6 +56,44 @@
     }
 }
 
+@mixin lumx-text-field-material-input-wrapper-validity($validity, $theme) {
```
matmkian: Useless in my mind, we never do validation on textarea.

src/components/text-field/style/lumapps/_index.scss
juanigalan91: `flex: 1` does not work well with resizing the text area. In browsers that do not support the `resize` css property, it allows changes like this:
Screenshot 2019-10-22 at 10 13 41

src/components/text-field/style/lumapps/_index.scss
juanigalan91: yes, I compared the css with the ones here:
http://ui.lumapps.com/components/text-field

src/components/text-field/react/TextField.tsx
```diff
@@ -77,6 +83,12 @@ const COMPONENT_NAME = `${COMPONENT_PREFIX}TextField`;
  */
 const CLASSNAME: string = getRootClassName(COMPONENT_NAME);
 
+/**
+ * The line height for the Text Area. Please keep in mind that if that changes via css, the prop lineHeight needs to be changed
```
gcornut: There is a library that can calculate the line height of any element dynamically: https://www.npmjs.com/package/line-height
@juanigalan91 Do you think it could be a good option here?

gcornut: It would be nice to test. Trying to modify the line height in css and check that this lib can calculate the correct line height with reasonable performance

juanigalan91: I was just looking for a way to avoid the hardcoding and apparently we can use `event.target.clientHeight`. I already tested it and seems to be working as expected. and maybe we can avoid the library :)

gcornut: Ok nice job :+1:

src/components/text-field/react/TextField.tsx
```diff
@@ -92,20 +104,60 @@ const DEFAULT_PROPS: Partial<TextFieldProps> = {
...
+             * on the state in order to allow React to re-render. Therefore, we save them using `useState`
+             */
+            (event.target as HTMLTextAreaElement).rows = MIN_ROWS;
+            const currentRows = event.target.scrollHeight / lineHeight;
+            (event.target as HTMLTextAreaElement).rows = currentRows;
+
+            setRows(currentRows);
+        }
+        onChange(get(event, 'target.value'));
+    };
```
gcornut: Could the compute of the number of rows be extracted in a separate hook that would depend on the textarea ref and produce the number of rows?
Something like:
```jsx
const rows = useComputeNumberOfRows(textAreaRef);
....
<textarea rows={rows}
```
The hook could be private to this component. It would use an internal `useState` to store the number of rows and a `useEffect` to listen to the on change event of the textarea (separatelly from this on change handler) to update the `rows` inner state

src/components/text-field/react/TextField.tsx
```diff
@@ -48,6 +48,12 @@ interface ITextFieldProps extends IGenericProps {
     /** Theme. */
     theme?: Theme;
 
+    /** Line height of the input or text area element. */
+    lineHeight?: number;
```
gcornut: Should this be replaced with the minimum number of rows since you now calculate the line height dynamically?

src/components/text-field/react/TextField.tsx
juanigalan91: Sure, great idea! I will take that approach :)

src/components/text-field/react/TextField.tsx
juanigalan91: yep, forgot to deprecate that one.

src/components/text-field/react/TextField.tsx
```diff
@@ -81,35 +87,106 @@ const COMPONENT_NAME = `${COMPONENT_PREFIX}TextField`;
...
     type: TextFieldType.input,
 };
-
+/**
+ * Hook that allows to calculate the number of rows needed for a text area.
+ * @param minimumNumberOfRows Minimum number of rows that we want to display.
+ * @return rows to be used and a callback to recalculate
+ */
+const useComputeNumberOfRows = (
+    minimumNumberOfRows: number | undefined = MIN_ROWS,
```
gcornut: It's probably easier to read as `minimumNumberOfRows?: number` rather than `minimumNumberOfRows: number | undefined`
Or alternativaly, we could ask the `minimumNumberOfRows` to be always provided in htis hook and use the default prop value make sure it's not `undefined` (we also would not need the `MIN_ROWS` const anymore)

gcornut: For the second option, you have to tell TypeScript that the `minimumNumberOfRows` cannot be `undefined` like this:
`minimumNumberOfRows = DEFAULT_PROPS.minimumNumberOfRows as number`

juanigalan91: For me, it feels more natural to avoid passing the prop and let the component figure out what to use. I will make the changes so it reads as `optional`

PR #208, opened 2019-10-25
feat(uploader): add component

matmkian:
# General summary

Add uploader component.

# Screenshots

Capture d’écran 2019-10-25 à 14 26 28


gcornut:
@matmkian I've done the React part. Here is what I added which is not in angular:
- Default size is set to XL since the component does not look good with sizes under that
- Size range is limited to XL and XXL
- The aspect ratio will be overriden to "square" if the variant is "circle" to prevent weird oval shape

Let me know if you are against any of theses. I can implement them in Angular too if you want

matmkian:
@gcornut, you're so perfect! If you can implement those changes in the Angularjs version that would be perfect!

FlorentDinet:
I see a lot of differences with the mockups https://app.zeplin.io/project/5beed1207b5f913ebc61d14e/dashboard?seid=5dadbbbc7ef2d22c514c286b
is this wanted ?

Also AspectRatio.original seems broken (no background and somewhat no bounding box)
Capture d’écran 2019-11-06 à 16 38 33
Capture d’écran 2019-11-06 à 16 38 58

matmkian:
Need to talk about that @FlorentDinet , only did https://app.zeplin.io/project/5beed1207b5f913ebc61d14e/screen/5dbfe0c71ad43e22ffd2e906 and no original aspect ratio

FlorentDinet:
Ok to discuss, it's a nice start let's merge, and improve later

PR #219, opened 2019-11-05
feat(tree-shake): tree shake lodash

juanigalan91:
# General summary
While taking a look at the generate bundle for production using `webpack-bundle-analyzer`, I noticed that we were importing the entire `lodash` dependency, but we are actually only using some modules:
Screenshot 2019-11-05 at 17 25 03

This was currently adding 20 KB (GZipped) to the final bundle that we are using in our product.

Changing the imports in order to allow tree shaking, we can see that the bundle decreased:
Screenshot 2019-11-05 at 17 25 10

Before tree shaking:
62.26 KB GZipped.

After tree shaking:
40.72 KB GZipped.

# Check list

### DIFF
--- a/src/components/comment-block/react/CommentBlock.tsx
+++ b/src/components/comment-block/react/CommentBlock.tsx
@@ -8,7 +8,7 @@ import { Avatar, Size, Theme } from 'LumX';
 import { ENTER_KEY_CODE } from 'LumX/core/constants';
 import { IGenericProps, getRootClassName } from 'LumX/core/react/utils';
 import { handleBasicClasses } from 'LumX/core/utils';
-import { isFunction } from 'lodash';
+import isFunction from 'lodash/isFunction';
 
 /////////////////////////////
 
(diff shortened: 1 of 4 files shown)


gcornut:
I've thought a bit about that. There is an alternative using a babel plugin:
https://github.com/lodash/babel-plugin-lodash

This plugin automatically rewrites lodash import so that you do grouped lodash import but still with tree shaking.
I didn't test though

juanigalan91:
thanks @gcornut ! it definitely looks prettier with that approach, I will give it a go!

juanigalan91:
@gcornut I have tried the plugin and I can't seem to make it work. here are my changes:
https://github.com/lumapps/design-system/pull/new/feat%2Ftree-shake-lodash

I am not that familiar with the webpack scaffolding, maybe I am adding the plugin on the wrong file 😅

gcornut:
@juanigalan91 Yeah it does not appear to work. Maybe it also needs the `lodash-webpack-plugin`...
This may be too much addons where we could just use the partial import like you did

I'd say we can go with your solution and later, when we will rework the project architecture and build we could give an other try.

Otherwise we could also add an [eslint rule](https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/import-scope.md) to enforce the use of these specific type of lodash import

juanigalan91:
> @juanigalan91 Yeah it does not appear to work. Maybe it also needs the `lodash-webpack-plugin`...
> This may be too much addons where we could just use the partial import like you did
> 
> I'd say we can go with your solution and later, when we will rework the project architecture and build we could give an other try.
> 
> Otherwise we could also add an [eslint rule](https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/import-scope.md) to enforce the use of these specific type of lodash import

The eslint rule seems like a great idea ! the plugin would be the best solution, but as you said, we can give it a go once the restructure is in place.

gcornut:
While we restructured the project, we modified the lodash imports to enable tree shaking so this PR is not needed anymore.
Thanks @juanigalan91  for pointing this out!

PR #237, opened 2019-11-26
feat(storybook): added storybook  💅🏻

juanigalan91:
# General summary
So, Storybook for DS  !

There were several decisions taken (which we definitely need to discuss) and some small changes:
- Storybook was installed directly on the `lumx/react` package. Why ? We could have installed storybook in the root folder but:
    - The idea with lerna is to manage dependencies for each package
    - Since we have different technologies (react, angular), if we wanted to use storybook for angular, we would need to install `@storybook/angular`, which makes more sense to install directly in the `lumx/angularjs` package and not in the root folder
- A quick `npm script` was added to execute storybook `yarn storybook:react` which moves to the `lumx/react` and executes the storybook.
- Storybook webpack config needs to reuse all the loaders that we have on the package itself, in order to properly load the files. That's why it imports the webpack config and extends it.
- Stories were written using [CSF](https://storybook.js.org/docs/formats/component-story-format/) since that is the new way for stories. Personally I am not a big fan of the API, but any new features will be added to CSF and not the old API so 🤷🏻‍♂️

Still pending:
- Add several addons to Storybook, like automatic documentation generation using Typescript
- Theme Switcher
- Maybe add storyshots ? 
- Start writing stories for each component 😅

# Screenshots
Screenshot 2019-11-26 at 15 42 57

# Check list

-   [X] (if has textual documentation) Add `need: review-doc` label

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,14 @@ yarn start
 
 You can then open your browser to [http://localhost:4000/](http://localhost:4000/) to access the demo and test site.
 
+You can also execute a Storybook to develop your components, in order to that you just need to execute:
+
+```bash
+yarn storybook:react
+```
+
+**Note**: storybook is currently configured for React components.
+
 ## Project build
 
 you can build LumX by using:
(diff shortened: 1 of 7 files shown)


Flo5k5:
Awesome!

gcornut:
Test OK

PR #238, opened 2019-11-26
fix(react): fix build typescript declaration file

gcornut:
# General summary

WIP fix the generation of the typescript declaration file (that should be located under `dist/lumx.react.d.ts`)

Currently in error: 
https://github.com/unimonkiez/ts-declaration-webpack-plugin/issues/4


gcornut:
Fixed in https://github.com/lumapps/design-system/pull/322

PR #248, opened 2019-11-28
feat(autocomplete): stories and fixes

juanigalan91:
# General summary
In this PR, the following changes were introduced:
- Autocomplete and AutocompleteMultiple stories
- When we are using the AutocompleteMultiple, there is a necessity to just display the suggestions with the same size as the input. This implies two changes:
    - Tell the Autocomplete component that sometimes we want to anchor the suggestions to the input rather than the wrapper
    - `useComputePosition` ignored the `hasParentWidth` parameter when setting the minimum width for the dropdown. This PR changes that in order to ensure that the minimum width is not set when we want to have the same width as the parent
- For the Autocomplete simple, we do not want to focus upon close, so we pass down a prop in order to signal that we do not want to apply the `useFocusOnClose` hook.
- This is all configurable via props, so if we want the old behaviour, we certainly can !

# Screenshots
8Tvz0XIN4P

# Check list

### DIFF
--- a/packages/lumx-react/src/components/autocomplete/Autocomplete.tsx
+++ b/packages/lumx-react/src/components/autocomplete/Autocomplete.tsx
@@ -15,6 +15,12 @@ import { useFocusOnClose } from '@lumx/react/hooks/useFocusOnClose';
  * Defines the props of the component.
  */
 interface IAutocompleteProps extends IGenericProps {
+    /**
+     * Whether the suggestions list should display anchored to the input
+     * If false, it will be anchored to the text field wrapper.
+     */
+    anchorToInput: boolean;
+
     /** A ref that will be passed to the input or textarea element. */
     inputRef?: RefObject<HTMLInputElement>;
 
@@ -41,6 +47,11 @@ interface IAutocompleteProps extends IGenericProps {
      * @see {@link TextFieldProps#hasError}
      */
     hasError?: boolean;
+
+    /**
+     * Whether the text box should be focused upon closing the suggestions
+     */
+    shouldFocusOnClose?: boolean;
 
     /**
      * Whether the text field displays a clear button or not.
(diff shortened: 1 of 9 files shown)


packages/lumx-react/src/components/autocomplete/Autocomplete.tsx
```diff
@@ -220,7 +235,7 @@ const Autocomplete: React.FC<AutocompleteProps> = (props: AutocompleteProps): Re
     } = props;
 
     const textFieldRef = useRef(null);
-    useFocusOnClose(inputRef.current, isOpen);
+    useFocusOnClose(inputRef.current, isOpen, shouldFocusOnClose);
```
gcornut: It may be a bit better to have the if check on `shouldFocusOnClose` here rather than in the `useFocusOnClose`. 
That way we wouldn't even need a hook if the `shouldFocusOnClose` boolean value is set to false

juanigalan91: yeah, I actually took that approach while I was coding, but I got a beautiful error from the linter reminding of the Rule of hooks, that they should always run and validate if they should be executed inside the hook. That's why I moved it to the hook. In any case, I have not seen issues related to that during my time using hooks, but on the other hand, it is discouraged by the React team. thoughts?

gcornut: Ok I imagine it's to make sure the hook is correctly removed when `shouldFocusOnClose` switches from `true` to `false`. It should not happen very often but you are right, we should keep it

FlorentDinet:
Excellent, and storybook for testing seems so much better !

I watched a minor thing, when pressing "tab" we lose the keyboard interaction with arrow keys:
Dec-06-2019 09-34-43

juanigalan91:
@FlorentDinet got it! will take a look at it right away :) thanks !

juanigalan91:
@FlorentDinet would you mind checking once more ? Thanks !

FlorentDinet:
ok you locked tab key when suggestions are open, hardcore but effective 👍 
not sure about accessibility though, but it's a good start.

👌 for me

PR #272, opened 2019-12-06
docs(language-switcher): doc text

FlorentDinet:
# General summary

Documentation for the language switcher component

Need dev-integration to replace demo blocks react + angular

# Screenshots

Capture d’écran 2019-12-06 à 16 02 17

# Check list

-   [x] (if has textual documentation) Add `need: review-doc` label


packages/site-demo/content/product/components/language-switcher/index.mdx
```diff
@@ -0,0 +1,25 @@
+# Language switcher
+
+**Language switchers display language options in a list and present the users choice. Use it when users need to select a language.**
```
lionelmoi:
```suggestion
**Language switchers display language options in a list and present the option the user chose. Use it when users need to select a language.**
```
It may seem a little heavy but it disambiguates.

PR #274, opened 2019-12-11
fix(demo): bring back demo grid lost during the great refactoring

pidupuis:


packages/site-demo/content/product/components/grid/index.mdx
```diff
@@ -0,0 +1,5 @@
+# Grid
+
+## Default
+
+<DemoBlock demo="default" />
```
gcornut: I don't see the angularjs/default.html file ^^
Also we did not account for angular demo without a controller I think. It might not be too difficult to allow this in the DemoBlock

pidupuis: The demo files are already in the repo (both the html and the controller)

gcornut: Ah ok

PR #281, opened 2019-12-19
Improve date-picker

pidupuis:
# General summary

- [x] Use Popover instead of Dropdown
- [x] Add simple tests
- [ ] Add keyboard interaction

# Check list


packages/lumx-react/src/components/date-picker/DatePicker.tsx
```diff
@@ -29,10 +29,10 @@ interface IDatePickerProps extends IGenericProps {
     todayOrSelectedDateRef?: RefObject<HTMLButtonElement>;
 
     /** Value. */
-    value: moment.Moment;
+    value: moment.Moment | undefined;
 
     /** On change. */
-    onChange(value: moment.Moment): void;
+    onChange(value: moment.Moment | undefined): void;
```
gcornut: Actually, re-thinking about it, it might be better to use `null` as the empty value

pidupuis: Not compatible with Moment isSame signature :/

PR #282, opened 2019-12-20
feat(icon): add has shape attribute

matmkian:
# General summary

# Screenshots

# Check list


pidupuis:
> Is that normal that for the lumapps theme in dark mode, somes icons are not visible ?
> Capture d’écran 2019-12-20 à 12 04 24
> 
> Capture d’écran 2019-12-20 à 12 04 38

Yes, it's WAD. Icon does not handle themes for now. And for the shape to be visible, we make a default color as dark.

One day we will handle theme and the default color will be light on dark theme. But for now it's WAD.

pidupuis:
icon_test

PR #288, opened 2019-12-24
feat(message): add Message component in react with tests

gigeos:
# General summary

Add Message component in React

# Screenshots

Capture d’écran 2019-12-24 à 17 30 15

# Check list

-   [x] Add `need: review-frontend` label
-   [ ] Check through the [react dev check list]
-   [ ] Add `need: test` label

### DIFF
--- a/packages/lumx-react/src/components/message/Message.tsx
+++ b/packages/lumx-react/src/components/message/Message.tsx
@@ -1,6 +1,6 @@
 import { mdiAlert, mdiAlertCircle, mdiCheckCircle, mdiInformation } from '@lumx/icons';
 import { ColorPalette, Icon } from '@lumx/react';
-import { COMPONENT_PREFIX, CSS_PREFIX } from '@lumx/react/constants';
+import { COMPONENT_PREFIX } from '@lumx/react/constants';
 import { IGenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
 import classNames from 'classnames';
 import React, { ReactElement, ReactNode } from 'react';
@@ -97,12 +97,10 @@ const Message: React.FC<MessageProps> = (props: MessageProps): ReactElement => {
             className={classNames(
                 className,
                 handleBasicClasses({
+                    color,
                     hasBackground,
                     prefix: CLASSNAME,
                 }),
-                {
-                    [`${CSS_PREFIX}-message--color-${color}`]: true,
-                },
             )}
             {...forwardedProps}
         >
--- a/packages/lumx-react/src/components/message/Message.tsx
+++ b/packages/lumx-react/src/components/message/Message.tsx
@@ -3,11 +3,11 @@ import { ColorPalette, Icon } from '@lumx/react';
 import { COMPONENT_PREFIX, CSS_PREFIX } from '@lumx/react/constants';
 import { IGenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
 import classNames from 'classnames';
-import React, { ReactElement } from 'react';
+import React, { ReactElement, ReactNode } from 'react';
 
 /////////////////////////////
 
-enum KindMessage {
+enum MessageKind {
     error = 'error',
     info = 'info',
     success = 'success',
@@ -18,6 +18,11 @@ enum KindMessage {
  * Defines the props of the component.
  */
 interface IMessageProps extends IGenericProps {
+    /**
+     * Message content.
+     */
+    children?: ReactNode;
+
     /**
      * The kind of message.
      */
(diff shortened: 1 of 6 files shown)


packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,125 @@
...
+ */
+const Message: React.FC<MessageProps> = (props: MessageProps): ReactElement => {
+    const { children, className, kind, hasBackground, ...forwardedProps } = props;
+    const icon = kind ? _KIND_ICON[kind] : null;
+    /**
+     * Get message classes.
+     *
+     * @return The list of message classes.
+     */
+    function getClasses(): string[] {
```
Flo5k5: I think you can use handleBasicClasses

packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,125 @@
+import { mdiAlert, mdiAlertCircle, mdiCheckCircle, mdiInformation } from '@lumx/icons';
+import React, { ReactElement } from 'react';
+
+import { ColorPalette, Icon } from '@lumx/react';
+import classNames from 'classnames';
+
+import { COMPONENT_PREFIX, CSS_PREFIX } from '@lumx/react/constants';
+import { IGenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
+
+/////////////////////////////
```
Flo5k5: Missing line return

packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,125 @@
...
+enum KindMessage {
+    error = 'error',
+    info = 'info',
+    success = 'success',
+    warning = 'warning',
+}
+/**
```
Flo5k5: Missing line return

packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,125 @@
+import { mdiAlert, mdiAlertCircle, mdiCheckCircle, mdiInformation } from '@lumx/icons';
+import React, { ReactElement } from 'react';
+
+import { ColorPalette, Icon } from '@lumx/react';
```
Flo5k5: Sort and group imports

packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    return (
+        <div
+            className={classNames(
+                className,
+                handleBasicClasses({
+                    prefix: CLASSNAME,
+                }),
+                getClasses(),
```
Flo5k5: I think you can use handleBasicClasses instead (same purpose)

packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * The kind of message.
+     */
+    kind?: string;
+
+    /**
+     * Put a background to the message
+     */
+    hasBackground?: boolean;
+}
```
gcornut: It would be nice to have the `children` props here to document it's use as the message content

packages/lumx-react/src/components/message/Message.tsx
gcornut: Yes indeed, you should be able to use 
```js
 handleBasicClasses({
                    prefix: CLASSNAME,
                    color: kind,
                    hasBackground
                }),
```
And you wouldn't need the `getClasses` function

packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,126 @@
...
+import React, { ReactElement } from 'react';
+
+/////////////////////////////
+
+enum KindMessage {
+    error = 'error',
+    info = 'info',
+    success = 'success',
+    warning = 'warning',
+}
```
gcornut: "KindMessage" makes it sound like the message is kind xD
Maybe we could re-name it "MessageKind"?

pidupuis: `NiceChat`, `PepTalk`, `InspirationQuote`, `WellSaid`

packages/lumx-react/src/components/message/Message.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <div
+            className={classNames(
+                className,
+                handleBasicClasses({
+                    hasBackground,
+                    prefix: CLASSNAME,
+                }),
+                {
+                    [`${CSS_PREFIX}-message--color-${color}`]: true,
+                },
```
gcornut: It wasn't working like this? =>
```js
handleBasicClasses({
  hasBackground,
  color,
  prefix: CLASSNAME,
})
```

packages/lumx-react/src/index.ts
```diff
@@ -48,6 +48,8 @@ export { ListItem, ListItemProps, ListItemSize, ListItemSizes } from './componen
 
 export { ListSubheader, ListSubheaderProps } from './components/list/ListSubheader';
 
+export { CLASSNAME, DEFAULT_PROPS, MessageKind, Message, MessageProps } from './components/message/Message';
```
gcornut: I think the `CLASSNAME` and `DEFAULT_PROPS` don't need to be exported here (at least we didn't find reasons to do it for other components).
If we need them to be exported, we would all need to rename them I think

gigeos: I export them for use in my tests. Do I have to remove exports and duplicate new `DEFAULT_PROPS` and `CLASSNAME` for tests?

gcornut: The tests can use these from the `components/message/Message.tsx` file directly rather than in this `index.ts` file

PR #293, opened 2020-01-08
feat(input-helper): add component

radiuso:
# General summary

The new "Input Helper" has appeared!
The goal of this component is to provide a general guideline regarding Input messages like help message and error message.

The TextField now includes this new component through props :

- error = the error message to display. It is rendered only if hasError
- helper = the help message to display. Always displayed

# Screenshots
Capture d’écran 2020-01-08 à 15 12 15

# Check list


packages/lumx-react/src/components/index.ts
```diff
@@ -92,6 +92,16 @@ enum AspectRatio {
     square = 'square',
 }
 
+/**
+ * Semantic info about the purpose of the component
+ */
+enum Kind {
```
juanigalan91: Do we have any other scenarios throughout DS where we are using a sort of `type` or `kind` for a component? I am asking in order to keep coherence in the repo. If we are already using `type` rather than `kind`, maybe it would be great to keep that naming convention :)

packages/lumx-react/src/components/input-helper/InputHelper.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import { InputHelper, Kind } from '@lumx/react';
+import { decorators } from '@lumx/react/story-block';
+import { select, text } from '@storybook/addon-knobs';
+import React from 'react';
+
+export default { title: 'Input Helper', decorators };
+
+export const simpleHelper = ({ theme }) => (
```
juanigalan91: maybe this is more of a `informationHelper` rather than a `simpleHelper` ? Since the `kind` is `info` ?

packages/lumx-react/src/components/text-field/TextField.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+import noop from 'lodash/noop';
+import React from 'react';
+
+export default { title: 'TextField', decorators };
+
+/**
+ * TextField story
+ * @return simple TextField.
+ */
+export const simpleTextField = ({ theme }) => (
```
juanigalan91: nice! thanks for adding stories for the Text field !!

packages/lumx-react/src/components/text-field/TextField.test.tsx
```diff
@@ -49,7 +56,12 @@ const setup = (props: ISetupProps = {}, shallowRendering: boolean = true): ISetu
 
     const textarea = wrapper.find('textarea');
     const input = wrapper.find('input');
+    const error = wrapper.findWhere((n) => n.name() === 'InputHelper' && n.prop('kind') === Kind.error);
```
juanigalan91: 💪🏻

packages/lumx-react/src/components/index.ts
gcornut: I agree that `Kind` might not be the best name. Something like `InfoLevel` would be better.
We need this enum in the Notification component, the Message component and the Helper component

packages/lumx-react/src/components/text-field/TextField.tsx
```diff
@@ -23,9 +23,15 @@ interface ITextFieldProps extends IGenericProps {
     /** A Chip Group to be rendered before the main text input */
     chips?: HTMLElement | ReactNode;
 
+    /**  */
+    error?: ReactNode;
```
gcornut: I would maybe put `string | ReactNode` as a type. Even though `string` is already part of the union type `ReactNode`.
I think it would encourage the user to use a string by default.
Also a JSdoc string would be nice ^^
PS: the same goes for the `helper` prop

packages/lumx-react/src/components/input-label/InputLabel.tsx
```diff
@@ -0,0 +1,48 @@
...
+import classNames from 'classnames';
+import React, { ReactElement } from 'react';
+
+/////////////////////////////
+
+/**
+ * Defines the props of the component.
+ */
+interface IInputLabelProps extends IGenericProps {
+    theme?: Theme;
```
gcornut: You can add the `children` prop for documentation purpose

packages/lumx-react/src/components/index.ts
radiuso: Not sure of what is best.
It's a bit strange to have `kind={InfoLevel.warning}`

For Notification and Message, the enum could be a different one, since they use `success` instead of `valid` for the `InputHelper` component.
This is a semantic difference ^^

CHANGELOG.md
```diff
@@ -9,6 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Added
+
+-   Added `helper` prop to `TextField` component
+-   Added `error` prop to `Autocomplete`, `Select` and `TextField` component
+-   Added `lx-helper` and `lx-error` props to `lx-text-field` and `lx-select` components
+
+### Changed
+
 -   _[BREAKING]_ Changed `type` prop to `multiline` flag prop in the `TextField` component
```
gcornut: There is two `### Changed` section in here! ^^

packages/lumx-react/src/components/index.ts
gcornut: If this enum is too specific to the `InputHelper` component, we can just put it in the component file and name it `InputHelperKind`

packages/lumx-react/src/components/index.ts
```diff
@@ -92,6 +92,16 @@ enum AspectRatio {
...
+    info = 'info',
+    valid = 'success',
```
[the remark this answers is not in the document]
radiuso: This is linked to the fact that we say "TextField" "isValid"
So the helper reflect this idea of valid info instead of success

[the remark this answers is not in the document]
gcornut: Hmm ok but I'd rather always use valid or always use success not a mixture of the two then :/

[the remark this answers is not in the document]
matmkian: I suggest a rework of those constants after merging -> we have to share those constants between message, notification and input helper components in both technologies (react / angularjs).
What do you think about that? @gcornut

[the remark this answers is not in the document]
gcornut: ok

PR #300, opened 2020-01-15
chore: update storybook and simplify its setup

gcornut:
# General summary

- Update storybook
- Simplify its configuration (https://medium.com/storybookjs/declarative-storybook-configuration-49912f77b78)
- Improve theme switcher
Capture d’écran de 2020-01-17 13-02-27

# Check list


gcornut:
@juanigalan91 let me know what you think of these improvements (and if you also have time to test these it would be nice ^^)

packages/lumx-react/.storybook/main.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
juanigalan91: looks awesome ! extremely simple :)

juanigalan91:
@gcornut I tried it out, works like a charm! Great work 💪🏻

However, I do feel that the themes (Material and Lumapps) and the light/dark are two different things, managed in different states and change different look and feel on the page. Even at a code level we are managing them separately, so from a developer experience, it feels a little bit weird to have these two things with the same UI and next to each other, since they look like one "radio button" group.

In any case, nothing blocking, I just wanted to point that out :)

matmkian:
I agree, at least a little label in front of chip groups like on the demo site
Capture d’écran 2020-01-16 à 08 59 33

gcornut:
> @gcornut I tried it out, works like a charm! Great work 💪🏻
> 
> However, I do feel that the themes (Material and Lumapps) and the light/dark are two different things, managed in different states and change different look and feel on the page. Even at a code level we are managing them separately, so from a developer experience, it feels a little bit weird to have these two things with the same UI and next to each other, since they look like one "radio button" group.
> 
> In any case, nothing blocking, I just wanted to point that out :)

Ok, I see. The main thing I wanted to acheive is to have both "theme" switcher on at the top so that hey both stay in place on every story. But yeah we could add labels and have the dark/light theme switcher use the `Switch` component like before

matmkian:
Test ko -> when i switch from a story to another, no more style :-(

matmkian:
Test KO -> les thèmes sont inversés et rien dans les stories d'autocomplete.

PR #322, opened 2020-02-10
Chore/improve ts experience on npm

gcornut:
# General summary

- Rework of the TS compiler (babel based) & config
- Exposed TS types on NPM (for @lumx/core, @lumx/icons & @lumx/react)


packages/lumx-core/src/js/utils.ts
```diff
@@ -17,10 +16,11 @@ import { ESCAPE_KEY_CODE } from './constants';
 /**
  * Enhance isEmpty method to also works with numbers.
  *
- * @param  {any}     value The value to check.
- * @return {boolean} Weither if the input value is empty or != 0.
+ * @param  value The value to check.
+ * @return Whether if the input value is empty or != 0.
```
pidupuis:
```suggestion
 * @return Whether the input value is empty or != 0.
```

packages/lumx-react/src/components/autocomplete/__mockData__/index.ts
```diff
@@ -1,3 +1,4 @@
+/* tslint:disable object-literal-sort-keys */
```
pidupuis: We could remove this rule for good :man_shrugging:

gcornut: I'd say we can do that in another PR in which we remove all old rules, remove tslint and install latest version of eslint with the new rules from lumapps-web

Flo5k5: Btw we could extract rules from lumapps-web and create a separate package we could reuse in every web repo

PR #325, opened 2020-02-12
Feat/focal point react

Flo5k5:
# General summary

Handle focal point on thumbnails.

# Screenshots

# Check list

### DIFF
--- a/packages/lumx-react/src/components/thumbnail/IFocus.ts
+++ b/packages/lumx-react/src/components/thumbnail/IFocus.ts
@@ -1,0 +1,5 @@
+// Credits: https://github.com/third774/image-focus/
+export interface IFocus {
+    x: number;
+    y: number;
+}
--- a/packages/lumx-react/src/components/thumbnail/IFocus.ts
+++ b/packages/lumx-react/src/components/thumbnail/IFocus.ts
@@ -1,5 +1,0 @@
-// Credits: https://github.com/third774/image-focus/
-export interface IFocus {
-    x: number;
-    y: number;
-}
--- a/packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
+++ b/packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useEffect, useRef } from 'react';
 
 import classNames from 'classnames';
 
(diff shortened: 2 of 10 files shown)


packages/lumx-react/src/utils/debounce.ts
```diff
@@ -0,0 +1,7 @@
+export function debounce(func: (...args: any) => any, debounceTime: number) {
```
gcornut: debounce exist dans lodash

packages/lumx-react/src/components/thumbnail/IFocus.ts
```diff
@@ -0,0 +1,5 @@
+// Credits: https://github.com/third774/image-focus/
+export interface IFocus {
```
gcornut: Could use a more generic name like `Point2D` (Some similar interface are probably used in `useComputePosition`)

packages/lumx-react/src/utils/assign.ts
```diff
@@ -0,0 +1,4 @@
+export const assign = (target: any, ...sources: any[]) => {
```
gcornut: Why not use `Object.assign()` ?

packages/lumx-react/src/components/thumbnail/IFocus.ts
gigeos: I think we must have a specific interface, indeed they are coordinates but limited from -1 to 1 so if we use specific interface we know there are limitations

packages/lumx-react/src/utils/assign.ts
Flo5k5: 👍

packages/lumx-react/src/utils/debounce.ts
Flo5k5: 👍

matmkian:
Je fais le test à la fin de la review. Je dois aussi supprimer du css useless. je ferais tout ça quand review ok.

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
```diff
@@ -148,12 +158,31 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
...
+                    <img
+                        ref={(f): FocusedImage | undefined => {
+                            return f
+                                ? new FocusedImage(f as ILumHTMLImageElement, {
+                                      debounceTime: 17,
+                                      focus,
+                                      updateOnWindowResize: true,
+                                  })
+                                : undefined;
+                        }}
```
gcornut: Could it be possible to have a static method `FocusImage.attach` that could be used like this:
```tsx
<img ref={FocusedImage.attach} data-focus-x={focus.x} data-focus-y={focus.y} />
```
Could the focus point be controlled with the `data` attributes?
Also I think it would be clearer to rename `focus` to `focusPoint` everywhere to not confuse the HTML focus with this image focus

PR #331, opened 2020-02-17
feat(web-bookmark): add WebBookmark React component

hchevallier:
# General summary

Add a new component `WebBookmark`.

# Screenshots
Expected : 
image
image

# Check list

### DIFF
--- a/packages/lumx-react/src/components/web-bookmark/WebBookmark.tsx
+++ b/packages/lumx-react/src/components/web-bookmark/WebBookmark.tsx
@@ -1,12 +1,11 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 
 import classNames from 'classnames';
 
-import { AspectRatio, Theme, Thumbnail } from '@lumx/react';
+import { Alignment, AspectRatio, ColorPalette, ColorVariant, Link, Size, Theme, Thumbnail } from '@lumx/react';
 
 import { COMPONENT_PREFIX } from '@lumx/react/constants';
 import { IGenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
-import { Alignment, Size } from '..';
 
 /////////////////////////////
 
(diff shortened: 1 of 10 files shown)


packages/lumx-react/src/components/web-bookmark/WebBookmark.tsx
```diff
@@ -0,0 +1,105 @@
...
+        | string
+        | {
+              __html: string;
+          };
+    /* The size variant of the web bookmark. */
+    size?: Size.regular | Size.big;
+    /* Theme. */
+    theme?: Theme;
+    /* Thumbnail image source */
+    thumbnail?: string;
```
gcornut: Does the component displays correctly without a thumbnail? Should it be required here?

packages/lumx-react/src/components/web-bookmark/WebBookmark.tsx
```diff
@@ -0,0 +1,105 @@
...
+    thumbnail = '',
+}) => {
+    return (
+        <a
+            target="_blank"
+            href={url}
+            className={classNames(className, handleBasicClasses({ prefix: CLASSNAME, theme, size }))}
+        >
+            <div className={`${CLASSNAME}__thumbnail`}>
+                <Thumbnail image={thumbnail} align={Alignment.center} aspectRatio={AspectRatio.horizontal} />
```
gcornut: Since is not required in prop, does this component appear correctly with an empty string as an image?
Shouldn't this whole thumbnail div be hidden if the thumbnail prop is empty?

packages/lumx-react/src/components/web-bookmark/WebBookmark.tsx
hchevallier: Yes it does, it just leaves an empty space. But maybe we should add a placeholder. @FlorentDinet ?

PR #334, opened 2020-02-18
feat(notification): prevent showing notification when !isOpen

juanigalan91:
# General summary
The notification component is visible for a split second even if `isOpen = false`. In order to prevent that, we render `null` when `isOpen = false`

# Screenshots
Before:
Feb-18-2020 09-38-39

After:
iFUs9CFIWU

# Check list

-   [X] (if fix or feature) Add `need: test` label

### DIFF
--- a/packages/lumx-react/src/hooks/useDelayedVisibility.tsx
+++ b/packages/lumx-react/src/hooks/useDelayedVisibility.tsx
@@ -1,15 +1,14 @@
 import { useEffect, useState } from 'react';
 
-const ANIMATION_DELAY = 400;
-
 /**
  * Returns true if the component is visible taking into account the component's
  * own visibility and the animations delay
  *
- * @param isComponentVisible Whether the component intends to be visible or not
+ * @param isComponentVisible Whether the component intends to be visible or not.
+ * @param transitionDuration time in ms that the transition takes for the specific component.
  * @return true if the component should be rendered
  */
-function useDelayedVisibility(isComponentVisible: boolean) {
+function useDelayedVisibility(isComponentVisible: boolean, transitionDuration: number) {
     // Delay visibility to account for the 400ms of CSS opacity animation.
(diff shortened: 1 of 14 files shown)


juanigalan91:
Nice catch! yep, I broke the animation. I think I will add that piece of code to a hook and use them on both sides, does that sound good to you @gcornut ?

gcornut:
> Nice catch! yep, I broke the animation. I think I will add that piece of code to a hook and use them on both sides, does that sound good to you @gcornut ?

My solution to manually wait for the CSS animation to end (using a timeout) might not be the best though. Maybe we could explore a hook that uses something like this:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/animationend_event
(I've never used this HTML event before so I don't know if it is what we want)

juanigalan91:
@gcornut yep, sounds cool! Unfortunately it is not supported for IE and Edge, if we use it, we still need to default to the `setTimeout`. Should we add it anyway ?

gcornut:
> @gcornut yep, sounds cool! Unfortunately it is not supported for IE and Edge, if we use it, we still need to default to the `setTimeout`. Should we add it anyway ?
> 
> https://caniuse.com/#feat=mdn-api_window_animationend_event

Hmm you are right, let's use the timeout since we will have to keep it for compatibility

packages/lumx-react/src/hooks/useDelayedVisibility.tsx
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState } from 'react';
+
+const ANIMATION_DELAY = 400;
```
matmkian: Animation delay is 400 for dialog but only 200 for notifications

juanigalan91: @matmkian added those changes! thanks for the review!

PR #336, opened 2020-02-19
feat(link): add React version of Link component

hchevallier:
# General summary

Add the React version of the `Link` component.
Only takes a color and a colorVariant, every other props is passed to the `a` tag.

# Screenshots

# Check list

### DIFF
--- a/packages/lumx-react/src/components/link/Link.stories.tsx
+++ b/packages/lumx-react/src/components/link/Link.stories.tsx
@@ -1,0 +1,39 @@
+import { ColorPalette, ColorVariant, Link } from '@lumx/react';
+import { boolean, select, text } from '@storybook/addon-knobs';
+import React from 'react';
+
+const colors = {
+    Blue: ColorPalette.blue,
+    Dark: ColorPalette.dark,
+    Green: ColorPalette.green,
+    Light: ColorPalette.light,
+    Primary: ColorPalette.primary,
+    Red: ColorPalette.red,
+    Secondary: ColorPalette.secondary,
+    Yellow: ColorPalette.yellow,
+};
+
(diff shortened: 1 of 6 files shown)


packages/lumx-react/src/components/link/Link.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export default { title: 'Link' };
+
+export const simpleLink = () => (
+    <Link
+        href={text('Href', 'https://google.com')}
+        target={boolean('target blank', false) && '_blank'}
+        color={select('Color', colors, ColorPalette.blue)}
+        colorVariant={select('Color Variant', colorVariants, ColorVariant.N)}
```
gcornut: To simplify (no need for the colors and colorVariants const):
```js
        color={select('Color', ColorPalette, ColorPalette.blue)}
        colorVariant={select('Color Variant', ColorVariant, ColorVariant.N)}
```

packages/lumx-react/src/components/link/Link.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Color, ColorVariant } from '@lumx/react';
+import { IGenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
+
+/////////////////////////////
+
+/**
+ * Defines the props of the component.
+ */
+
+interface ILinkProps extends IGenericProps {
```
gcornut: Since this component is really a wrapper around `<a>`, I think it would be nice to make the props extends the native anchor props.
```suggestion
interface ILinkProps extends HTMLAttributes<HTMLAnchorElement> {
```
The IGenericProps would not be necessary I think.

PR #337, opened 2020-02-19
feat(select): allow on Blur

juanigalan91:
# General summary
Added the possibility to detect whether the `Select` component was blurred or not:
- Since the Select is actually not an input, we need to fake the `blur` with a click away
- Updated the main story since it did not change the option selected upon clicking the option

# Check list

### DIFF
--- a/packages/lumx-react/src/components/select/Select.tsx
+++ b/packages/lumx-react/src/components/select/Select.tsx
@@ -304,21 +304,6 @@ const Select: React.FC<SelectProps> = ({
         [selectRef],
     );
 
-    // Any click away from the dropdown container will close it.
-    useClickAway(
-        selectRef,
-        () => {
-            if (!onBlur) {
-                return;
-            }
-
-            if (!isOpen && !isFocus) {
-                onBlur();
-            }
-        },
-        [selectRef],
-    );
-
     const handleKeyboardNav = useCallback(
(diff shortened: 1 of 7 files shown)


packages/lumx-react/src/components/select/Select.tsx
```diff
@@ -249,13 +268,47 @@ const Select: React.FC<SelectProps> = ({
...
+        setWasBlurred(false);
+    };
+
     useFocus(anchorRef.current, Boolean(isOpen));
-    useHandleElementFocus(anchorRef.current, setIsFocus);
+    useHandleElementFocus(anchorRef.current, setIsFocus, onSelectFocus);
+
+    /**
+     * Since the select is not really an input, we need to fake the onBlur event.
+     * We can do this by listening to the click away event, which is basically an "onBlur"
```
pidupuis: Smart

matmkian:
Test ko, need to demonstrate but when dropdown is open, if you click on another input, blur is triggered on second click.

juanigalan91:
@matmkian seems that this is an error that we have in master as well, since upon clicking on another input, the select is still focused.

In any case, I will take a look at the bug!

matmkian:
@juanigalan91 you're right, I just checked :-) Thank you!

matmkian:
@juanigalan91 Also, onBlur event is not triggered if you leave the field with tab key down.

PR #343, opened 2020-03-02
Fix/LUM-8316: split to unique and multiple

radiuso:
# General summary

Split the Select component in 2 main components

* Select -> string value
* SelectMultiple -> array as value

In order to avoid too many code duplication, an higher-order component "WithSelectContext" was created.

We can continue to split select if needed and create ChipSelect and ChipSelectMultiple.

## Important change

All unhandled props are now passed to the field instead of the wrapper.
The reason is, when we use a Select we expect to use a field. So if "disabled" is passed, it should go to the input and not the wrapper.

# Screenshots

# Check list


FlorentDinet:
Just demo issue I guess (the PR on left, master on right)

chipm
selectsearchm
selectm

packages/lumx-react/src/components/select/Select.tsx
```diff
@@ -238,233 +49,128 @@ const stopPropagation = (evt: Event) => evt.stopPropagation();
...
-                                    theme={theme}
-                                    onClick={onClear}
-                                    onKeyDown={stopPropagation}
-                                />
-                            )}
-
-                            <div className={`${CLASSNAME}__input-indicator`}>
-                                <Icon icon={mdiMenuDown} size={Size.s} />
-                            </div>
+}): React.ReactElement => {
```
gcornut: The return type here is not required since it is provided in the `React.FC` type signature

packages/lumx-react/src/components/select/SelectMultiple.tsx
```diff
@@ -0,0 +1,191 @@
...
+    },
+    selectedValueRender: (choice) => choice,
+};
+
+/**
+ * Select Multiple component.
+ *
+ * @return The component.
+ */
+const SelectMultiple: React.FC<any> = ({
```
gcornut: `React.FC<any>` really means this component accept any as props :/

packages/lumx-react/src/components/select/SelectMultiple.tsx
```diff
@@ -0,0 +1,191 @@
...
+    theme,
+    placeholder,
+    handleKeyboardNav,
+    targetUuid,
+    anchorRef,
+    isRequired,
+    isDisabled,
+    selectedChipRender = DEFAULT_PROPS.selectedChipRender,
+    selectedValueRender = DEFAULT_PROPS.selectedValueRender,
+}): React.ReactElement => {
```
gcornut: Same here return type not required

packages/lumx-react/src/components/select/Select.tsx
```diff
@@ -238,233 +49,128 @@ const stopPropagation = (evt: Event) => evt.stopPropagation();
...
+                </Chip>
             )}
-        </div>
+        </>
     );
 };
-Select.displayName = COMPONENT_NAME;
 
-export { CLASSNAME, DEFAULT_PROPS, Select, SelectProps, SelectVariant };
+const SelectWithContext = (props: SelectProps) => {
```
gcornut: I'm not a fan of this name. Could this component be just `Select`? the other component could be `SelectField` or `SelectInput`
 (It also goes for the SelectMultiple)

radiuso: Yes totally agree!
I changed the name for 'SelectField'

PR #345, opened 2020-03-02
Add ReactNode as an allowed type for SideNavigationItem

aureldent:
# General summary

Add ReactNode as an allowed type for SideNavigationItem to allow for component as labels.
This is only for typescript support to avoid hacky workarounds.

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   New component in React: `Mosaic`
 -   Added `size` prop to `WebBookmark` component
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


gcornut:
> Passing the node as children isn't enough?

The children are used to display sub menu (shown if `isOpen` is true) under the `SideNavigationItem` whereas the label is used in navigation item label itself. 
This modification is purely on TS type since string and JSX elements are both valid values where the label is currently used

Flo5k5:
> > Passing the node as children isn't enough?
> 
> The children are used to display sub menu (shown if `isOpen` is true) under the `SideNavigationItem` whereas the label is used in navigation item label itself.
> This modification is purely on TS type since string and JSX elements are both valid values where the label is currently used

Okay :)

PR #360, opened 2020-03-16
fix(thumbnail): fix internet explorer and issue if aspect ratio changes

gigeos:
# General summary

This should fix a display thumbnail problem on IE and size problem if you pass `aspectRatio` to `original` and then to an other value

# Screenshots

# Check list

### DIFF
--- a/packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
+++ b/packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
@@ -126,8 +126,24 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
     onClick = null,
     focusPoint = DEFAULT_PROPS.focusPoint,
     ...props
-}) => {
-    const focusImageRef = useFocusedImage(focusPoint!);
+}: ThumbnailProps): ReactElement => {
+    const focusImageRef = useFocusedImage(focusPoint!, aspectRatio!, size!);
+
+    /**
+     * Check if browser is IE
+     * @return Browser is IE or not
+     */
+    const isInternetExplorer = () => {
+        const ua = window.navigator.userAgent;
+        const msie = ua.indexOf('MSIE ');
+        const isIEVersion = !!navigator.userAgent.match(/Trident.*rv\:11\./);
+
+        if (msie > 0 || isIEVersion) {
+            return true;
+        }
+
+        return false;
+    };
 
     return (
         <div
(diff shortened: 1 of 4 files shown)


packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
```diff
@@ -139,7 +139,23 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
...
+     * @return Browser is IE or not
+     */
+    const isInternetExplorer = () => {
+        const ua = window.navigator.userAgent;
+        const msie = ua.indexOf('MSIE ');
+        const isIEVersion = !!navigator.userAgent.match(/Trident.*rv\:11\./);
+
+        if (msie > 0 || isIEVersion) {
+            return true;
+        }
```
juanigalan91: could we just do `return msie > 0 || isIEVersion` ? just to make it a little bit more simpler

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
```diff
@@ -139,7 +139,23 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
...
     ...props
 }: ThumbnailProps): ReactElement => {
-    const focusImageRef = useFocusedImage(focusPoint!);
+    const focusImageRef = useFocusedImage(focusPoint!, aspectRatio!, size!);
+
+    /**
+     * Check if browser is IE
+     * @return Browser is IE or not
+     */
+    const isInternetExplorer = () => {
```
gcornut: It could be very useful to have this function standalone in a util file. 
Something like: `src/utils/isInternetExplorer.ts`

packages/lumx-react/src/utils/isInternetExplorer.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Check if browser is IE
+ * @return Browser is IE or not
+ */
+const isInternetExplorer = () => {
+    const ua = window.navigator.userAgent;
+    const msie = ua.indexOf('MSIE ');
+    const isIEVersion = !!navigator.userAgent.match(/Trident.*rv\:11\./);
+    return msie > 0 || isIEVersion;
+};
+
```
juanigalan91: Since we will be using this function as an utils, could we add a little bit of unit testing to the function?

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
```diff
@@ -144,13 +146,13 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
...
             {aspectRatio === AspectRatio.original ? (
-                <img className={`${CLASSNAME}__image`} src={image} alt={alt} loading={loading} />
+                <img ref={focusImageRef} className={`${CLASSNAME}__image`} src={image} alt={alt} loading={loading} />
             ) : (
                 <div className={`${CLASSNAME}__background`}>
                     <img
                         ref={focusImageRef}
                         className={`${CLASSNAME}__focused-image`}
-                        crossOrigin="anonymous"
+                        crossOrigin={isInternetExplorer ? undefined : 'anonymous'}
```
gcornut: `isInternetExplorer` is a function, right?
```suggestion
                        crossOrigin={isInternetExplorer() ? undefined : 'anonymous'}
```

PR #363, opened 2020-03-17
fix(textfield-value): add the number as the value type

radiuso:
# General summary

The idea is to be able to handle number type in the TextField component

The onChange accept `any` as a type because it's more simple for the moment.
Maybe we can create a TextFieldValue type that will be string | number

# Screenshots

# Check list

### DIFF
--- a/packages/lumx-react/src/components/text-field/TextField.tsx
+++ b/packages/lumx-react/src/components/text-field/TextField.tsx
@@ -77,7 +77,7 @@ interface TextFieldProps extends GenericProps {
     textFieldRef?: RefObject<HTMLDivElement>;
 
     /** Text field value change handler. */
-    onChange(value: any): void;
+    onChange(value: string): void;
 
     /** Text field focus change handler. */
     onFocus?(event: React.FocusEvent): void;
@@ -173,7 +173,7 @@ interface InputNativeProps {
     rows: number;
     setFocus(focus: boolean): void;
     recomputeNumberOfRows(event: React.ChangeEvent): void;
-    onChange(value: any): void;
+    onChange(value: string): void;
     onFocus?(value: React.FocusEvent): void;
     onBlur?(value: React.FocusEvent): void;
 }
--- a/packages/lumx-react/src/components/text-field/TextField.tsx
+++ b/packages/lumx-react/src/components/text-field/TextField.tsx
@@ -71,13 +71,13 @@ interface TextFieldProps extends GenericProps {
     inputRef?: RefObject<HTMLInputElement> | RefObject<HTMLTextAreaElement>;
 
     /** Text field value. */
-    value: string;
+    value: string | number;
 
     /** A ref that will be passed to the wrapper element. */
     textFieldRef?: RefObject<HTMLDivElement>;
 
     /** Text field value change handler. */
-    onChange(value: string): void;
+    onChange(value: any): void;
 
     /** Text field focus change handler. */
     onFocus?(event: React.FocusEvent): void;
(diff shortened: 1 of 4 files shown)


packages/lumx-react/src/components/text-field/TextField.tsx
```diff
@@ -169,11 +169,11 @@ interface InputNativeProps {
     multiline?: boolean;
     maxLength?: number;
     placeholder?: string;
-    value: string;
+    value: string | number;
     rows: number;
     setFocus(focus: boolean): void;
     recomputeNumberOfRows(event: React.ChangeEvent): void;
-    onChange(value: string): void;
+    onChange(value: string | number): void;
```
pidupuis: Not `any` here either?

packages/lumx-react/src/components/text-field/TextField.tsx
```diff
@@ -71,13 +71,13 @@ interface TextFieldProps extends GenericProps {
     inputRef?: RefObject<HTMLInputElement> | RefObject<HTMLTextAreaElement>;
 
     /** Text field value. */
-    value: string;
+    value: string | number;
```
gcornut: I think we need this to be optional since for `number` value you may need to initialize the input as empty and the empty string is not very compatible with `number`
```suggestion
    value?: string | number;
```

pidupuis: Should be optional also

packages/lumx-react/src/components/text-field/TextField.tsx
```diff
@@ -71,13 +71,13 @@ interface TextFieldProps extends GenericProps {
...
 
     /** A ref that will be passed to the wrapper element. */
     textFieldRef?: RefObject<HTMLDivElement>;
 
     /** Text field value change handler. */
-    onChange(value: string): void;
+    onChange(value: any): void;
```
gcornut:
```suggestion
    onChange(value: string | number): void;
```
??
The problem is that the native input value is always string when accessed so I guess we need to do a `parseFloat` of the native value if the input type is `number`

pidupuis: How do you handle the following?
```js
const [foo, setFoo] = useState('foo');

// ...

    onChange={setFoo}

// ...
```

The signature are not the same.

I find the following quite ugly:
`onChange={v => setFoo(v as string)}`

gcornut: It's possible if we change the TextFieldProps to do something like:
```tsx
type TextFieldProps =
  | { type: 'number'; value?: number; onChange(value: number): void; }
  | { value?: string; onChange(value: string): void; }
```
^^
It could be the occasion to also do something like:

```tsx
type TextFieldProps =
  | ( { multiline: true } & TextareaHTMLAttributes<HTMLTextAreaElement>)
  | ( { multiline?: false } & InputHTMLAttributes<HTMLInputElement>)
```
So that the native props depends on the actual native input used

gcornut: It's possible if we change the TextFieldProps to do something like:
```tsx
type TextFieldProps =
  | { type: 'number'; value?: number; onChange(value: number): void; }
  | { value?: string; onChange(value: string): void; }
```
^^
It could be the occasion to also do something like:

```tsx
type TextFieldProps =
  | ( { multiline: true } & TextareaHTMLAttributes<HTMLTextAreaElement>)
  | ( { multiline?: false } & InputHTMLAttributes<HTMLInputElement>)
```
So that the forwarded props depends on the actual native input used

radiuso: I didn't succeed in this kind of implementation. 
I had errors with the onChange function.
Can we postpone these changes in another PR in order to give us some time to see exactly how to do these improvements?

gcornut: Ok but actually, now that I'm thinking about it the `onChange` still returns a value as a `string` rven if you give a `value` as a `number` because the native input value is always a `string`. 
So either we do a `parseFloat` when `type==='number'` (and keep `onChange(value: any): void`) or we keep the type `onChange(value: string): void;` 
What do you think @radiuso ?

radiuso: Yes, good point.
I think we can keep the `string` type. Maybe we want to use an integer or a float, the choice should not be done by the component.

gcornut: Ok
But parseFloat works with integer as well as float number ^^

PR #367, opened 2020-03-31
feat(link): add link ref on Link React component

gcornut:
# General summary

Add ref on Link component

### DIFF
--- a/packages/lumx-react/src/components/link/Link.test.tsx
+++ b/packages/lumx-react/src/components/link/Link.test.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { createRef, ReactElement } from 'react';
 
 import { mount, shallow } from 'enzyme';
 import 'jest-enzyme';
@@ -58,7 +58,11 @@ describe(`<${Link.displayName}>`, () => {
 
     // 2. Test defaultProps value and important props custom values.
     describe('Props', () => {
-        // Nothing to do here.
+        it('should use ref', () => {
+            const linkRef = createRef<HTMLAnchorElement>();
+            const { wrapper } = setup({ linkRef }, false);
+            expect(linkRef.current).toBe(wrapper.getDOMNode());
+        });
     });
 
     // 3. Test events.
--- a/packages/lumx-react/src/components/link/Link.tsx
+++ b/packages/lumx-react/src/components/link/Link.tsx
@@ -1,4 +1,4 @@
-import React, { AnchorHTMLAttributes } from 'react';
+import React, { Ref } from 'react';
 
 import { COMPONENT_PREFIX } from '@lumx/react/constants';
 import classNames from 'classnames';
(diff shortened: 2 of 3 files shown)


packages/lumx-react/src/components/link/Link.test.tsx
```diff
@@ -46,8 +46,10 @@ describe(`<${Link.displayName}>`, () => {
     // 1. Test render via snapshot (default states of component).
     describe('Snapshots and structure', () => {
         it('should render correctly', () => {
-            const { wrapper } = setup();
+            const linkRef = createRef<HTMLAnchorElement>();
+            const { wrapper } = setup({ linkRef });
             expect(wrapper).toMatchSnapshot();
+            expect(linkRef).toBeTruthy();
```
radiuso, line 52: Should we test that the link has the prop ref instead?

packages/lumx-react/src/components/link/Link.tsx
```diff
@@ -32,9 +35,13 @@ const CLASSNAME: string = getRootClassName(COMPONENT_NAME);
...
  * @return The component.
  */
-const Link: React.FC<LinkProps> = ({ children, className, color, colorVariant, ...props }) => {
+const Link: React.FC<LinkProps> = ({ children, className, linkRef, color, colorVariant, ...props }) => {
     return (
-        <a className={classNames(className, handleBasicClasses({ prefix: CLASSNAME, color, colorVariant }))} {...props}>
+        <a
+            className={classNames(className, handleBasicClasses({ prefix: CLASSNAME, color, colorVariant }))}
+            ref={linkRef}
+            {...props}
```
radiuso, line 43: Why not keeping the ref in the other props?

packages/lumx-react/src/components/link/Link.test.tsx
gcornut: Oh actually what I meant to test is the `linkRef.current` value which would indicate that the ref has been passed to the anchor and then the current value was updated.
I'm not sure which is best. Testing if the ref is passed to the anchor or testing is if the ref current value is update with the HTMLAnchorElement DOM object :thinking:

packages/lumx-react/src/components/link/Link.tsx
gcornut: Functional component cannot forward ref while forwarding props.
One way you can forward ref is to use `React.fowardRef` but it adds a wrapper around your functional component and for LumX we decided to use explicit name for Ref props everywhere.

radiuso: Good to know. 
Even if the `a` is primary element returned?

packages/lumx-react/src/components/link/Link.test.tsx
radiuso: Maybe it's equivalent in terms of what we test. 
But I guess that the responsibility of the component is to give the ref to the proper element

gcornut: I tried testing the "ref" prop as you say but it seems to be a special prop not directly accessible via the ezyme wrapper. 
So I decided to test that the linkRef.current was correctly filled with the link DOM object reference.
Let me know what you think of this!

packages/lumx-react/src/components/link/Link.test.tsx
```diff
@@ -58,7 +58,11 @@ describe(`<${Link.displayName}>`, () => {
 
     // 2. Test defaultProps value and important props custom values.
     describe('Props', () => {
-        // Nothing to do here.
+        it('should use ref', () => {
+            const linkRef = createRef<HTMLAnchorElement>();
+            const { wrapper } = setup({ linkRef }, false);
+            expect(linkRef.current).toBe(wrapper.getDOMNode());
+        });
```
gcornut, line 65: @radiuso I've perfected it ^^. What do you think?

PR #372, opened 2020-04-03
feat(side-navigation-item-link): add props to link in side nav item

juanigalan91:
# General summary
- Added a prop `linkProps` for the `SideNavigationItem` in order to provide additional props to the `a`, more importantly the `href` prop
- The prop was created in order to avoid confusions and mix up with the props related to the `li` on the top level, and differentiate where we want those props to go.
- For some reason I could not add a DEFAULT_PROP for this component, since tests were broken when I added to the `DEFAULT_PROPS` object a prop `linkProps: {}`. See photo in screenshots section

# Screenshots
Screenshot 2020-04-03 at 10 55 51

# Check list

### DIFF
--- a/packages/lumx-react/src/components/side-navigation/SideNavigationItem.tsx
+++ b/packages/lumx-react/src/components/side-navigation/SideNavigationItem.tsx
@@ -39,6 +39,9 @@ interface SideNavigationItemProps extends GenericProps {
     /** Whether or not the menu is selected. */
     isSelected?: boolean;
 
+    /** props that will be passed on to the Link */
+    linkProps?: Record<string, string>;
+
     /** On click handler. */
     onClick?(evt: React.MouseEvent): void;
 }
@@ -72,6 +75,7 @@ const SideNavigationItem: React.FC<SideNavigationItemProps> = (props) => {
         isOpen = DEFAULT_PROPS.isOpen,
         isSelected = DEFAULT_PROPS.isSelected,
         onClick,
+        linkProps,
         ...otherProps
     } = props;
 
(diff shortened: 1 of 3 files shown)


packages/lumx-react/src/components/side-navigation/SideNavigationItem.tsx
```diff
@@ -94,6 +98,7 @@ const SideNavigationItem: React.FC<SideNavigationItemProps> = (props) => {
                 tabIndex={0}
                 onClick={onClick}
                 onKeyDown={onClick ? onEnterPressed(onClick as Callback) : undefined}
+                {...(linkProps || {})}
```
gcornut, line 101: I think `{...(linkProps || {})}` should appear in first position to make sure it does not override the other props on `the <a>` element

packages/lumx-react/src/components/side-navigation/SideNavigationItem.tsx
```diff
@@ -39,6 +39,9 @@ interface SideNavigationItemProps extends GenericProps {
     /** Whether or not the menu is selected. */
     isSelected?: boolean;
 
+    /** props that will be passed on to the Link */
+    linkProps?: Record<string, string>;
```
gcornut, line 43: We can be more precise and type this as `React.DetailedHTMLProps<React.AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>`.
It's the official type of html anchor elements. It's quite complicated, I don't know if it should be used or not but it would add intellisense on the linkProps

juanigalan91:
@gcornut would you mind reviewing this PR one more time? I would love to have this feature for the new back office application

PR #381, opened 2020-04-21
feat(dialog): only use css for scrollable dialog

matmkian:
# General summary

# Screenshots

# Check list


packages/lumx-react/src/components/dialog/Dialog.tsx
```diff
@@ -185,11 +174,18 @@ const Dialog: React.FC<DialogProps> = (props) => {
...
+        () => {
+            if (preventAutoClose) {
+                return;
+            }
+
             onClose?.();
-        }
-    };
+        },
+        [],
```
gcornut, line 186: This array is optional. You can remove it

PR #384, opened 2020-04-22
Add custom default month for date picker

pidupuis:
# General summary
Add custom default month for date picker.

Fix https://lumapps.atlassian.net/browse/LUM-9212

Peek 23-04-2020 10-05
Peek 23-04-2020 10-06

# Check list


packages/lumx-react/src/components/date-picker/DatePicker.test.tsx
```diff
@@ -59,7 +59,7 @@ describe(`<${DatePicker.displayName}>`, () => {
     // 1. Test render via snapshot (default states of component).
     describe('Snapshots and structure', () => {
         it('should render correctly', () => {
-            const { wrapper } = setup();
+            const { wrapper } = setup({}, false);
```
gcornut, line 62: I think it's better to keep the shallow rendering. The snapshot for the `DatePicker` component should only contain the `DatePickerControlled` because `DatePicker` is a wrapper around it. 
Then, in the `DatePickerControlled.tests.js` you should have an other snapshot test with shallow rendering.
Sometimes shallow rendering is a bit finicky :/ but I really think there is great value in reviewing shallow snapshot that are focused on the tested component and does not dump the whole DOM tree ^^

PR #385, opened 2020-04-22
fix(infinite-scroll): trigger infinite when container is fully displayed

hchevallier:
# General summary
For autocompletes and select with infinite scroll, when there is enough place on screen to display the first batch of results, making the container unscrollable, it becomes impossible to trigger the infinite scroll.
The fix : When the container is displayed we check if the last item is visible, if so, we call the infinite scroll callback. 

# Screenshots

# Check list


pidupuis:
Test OK

PR #387, opened 2020-04-30
DatePicker and DPField accept a Date object

maximevast:
# General summary

Allow `<DatePicker />` and `<DatePickerField />` to accept a javascript Date object as property.

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
--   The `<DatePicker />` and `<DatePickerFields />` components now accept a javascript Date object as value.
+-   The `DatePicker` and `DatePickerFields` components now accept a javascript Date object or string as value.
 
 ## [0.22.0][] - 2020-04-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Added `type="button"` to `TextField` "clear" button when `isClearable` is `true` to avoid clearing field when user tries to submit a form.
 -   Moved `useFocusedImage` to `hooks` folder.
 
+### Changed
+
+-   The `<DatePicker />` and `<DatePickerFields />` components now accept a javascript Date object as value.
+
 ## [0.22.0][] - 2020-04-21
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   `TextField` component is now computing the number of rows (if `multiple` is set) at the first rendering
 
+### Changed
+
+-   The `DatePicker` and `DatePickerFields` components now accept a javascript Date object or string as value.
+
 ## [0.23.0][] - 2020-05-29
 
 ### Added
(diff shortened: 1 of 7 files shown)


gcornut:
@pidupuis Does it seem ok to you?
Also we need to test the compatibility with your PR https://github.com/lumapps/design-system/pull/384

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Trigger the infinite scroll callback on display if enough space with the `useInfiniteScroll` hook.
 
+### Changed
+
+-   The `<DatePicker />` and `<DatePickerFields />` components now accept a javascript Date object as value.
```
pidupuis, line 16:
```suggestion
-   The `DatePicker` and `DatePickerFields` components now accept a javascript Date object or string as value.
```

packages/lumx-react/src/components/date-picker/DatePicker.tsx
```diff
@@ -27,7 +27,7 @@ interface DatePickerProps extends GenericProps {
     todayOrSelectedDateRef?: RefObject<HTMLButtonElement>;
 
     /** Value. */
-    value: moment.Moment | undefined;
+    value: Date | moment.Moment | undefined;
```
pidupuis, line 30: Should also accept a compatible string.
Compatible string format is documented here: https://momentjscom.readthedocs.io/en/latest/moment/01-parsing/02-string/

packages/lumx-react/src/components/date-picker/DatePicker.tsx
```diff
@@ -57,7 +57,7 @@ const DEFAULT_PROPS: Partial<DatePickerProps> = {
  * @return The component.
  */
 const DatePicker = (props: DatePickerProps) => {
-    const today = props.value || moment();
+    const today = moment(props.value);
```
pidupuis, line 60: We have to be sure that `today` equals `moment()` if the value is `undefined` or if the string value is not parseable.

packages/lumx-react/src/components/date-picker/DatePickerField.test.tsx
```diff
@@ -49,19 +50,24 @@ const setup = ({ ...propsOverrides }: SetupProps = {}, shallowRendering: boolean
...
+        withDateObjectWrapper,
     };
 };
 
 describe(`<${DatePickerField.displayName}>`, () => {
     // 1. Test render via snapshot (default states of component).
     describe('Snapshots and structure', () => {
         it('should render correctly', () => {
-            const { wrapper } = setup();
+            const { wrapper, withDateObjectWrapper } = setup();
```
pidupuis, line 68: This is not how we test our component. You should create another dedicated `it()` block and use the `setup()` method to override the value. You should not create a new wrapper.

Something like this:
```js
const { wrapper } = setup({ value: new Date('January 18, 1970') });
```

packages/lumx-react/src/components/date-picker/DatePickerField.tsx
```diff
@@ -28,7 +28,7 @@ interface DatePickerFieldProps extends GenericProps {
     minDate?: Date;
 
     /** Value. */
-    value: moment.Moment | undefined;
+    value: Date | moment.Moment | undefined;
```
pidupuis, line 31: Should also handle compatible string. Since this type begins to be a bit complicated and is used inside different files, we should create a dedicated type and import it in the different files.

maximevast:
> @maximevast Could you make the two merge commit disappear the commit history?
> We prefer having clean commit history in PRs (by rebasing onto master rather than merging master into the branch)

Done :)

PR #391, opened 2020-05-07
feat(thumbnail): Img props

juanigalan91:
# General summary
- From the new Search UI, we want to add fallbacks to our images, since we have several scenarios where the images that we show are broken/not found.
- Therefore, we want to add a prop to Thumbnail that passes additional attributes to the `img` tag, so we can add, for example, an `onError`.

# Check list

### DIFF
--- a/packages/lumx-react/src/components/thumbnail/Thumbnail.stories.tsx
+++ b/packages/lumx-react/src/components/thumbnail/Thumbnail.stories.tsx
@@ -49,3 +49,44 @@ export const defaultThumbnail = ({ theme }: { theme: Theme }) => {
         />
     );
 };
+
+export const defaultThumbnailWithCustomImgProps = ({ theme }: { theme: Theme }) => {
+    const [image, setImage] = React.useState('https://not-found');
+    const onError = () => {
+        setImage('https://i.picsum.photos/id/1001/2400/1400.jpg');
+    };
+
+    return (
+        <Thumbnail
+            align={select<Alignment>('Alignment', Alignment, Alignment.left, 'Options')}
+            aspectRatio={select<AspectRatio>('Aspect ratio', AspectRatio, AspectRatio.square, 'Options')}
+            isCrossOriginEnabled={boolean('Enable CORS', true, 'Options')}
+            crossOrigin={select('CORS', CrossOrigin, CrossOrigin.anonymous, 'Options')}
+            fillHeight={boolean('Fill Height', false, 'Options')}
+            focusPoint={{
+                x: number('focusX', 0, numberKnobOtions, 'Options'),
+                y: number('focusY', 0, numberKnobOtions, 'Options'),
+            }}
+            image={image}
+            size={select(
+                'Size',
+                {
+                    XXS: Size.xxs,
+                    XS: Size.xs,
+                    S: Size.s,
+                    M: Size.m,
+                    L: Size.l,
+                    XL: Size.xl,
+                    XXL: Size.xxl,
+                },
+                Size.xxl,
+                'Options',
+            )}
+            theme={theme}
+            isFollowingWindowSize={boolean('Update on window resize', true, 'Options')}
+            resizeDebounceTime={number('Debounce time after resize', 20, undefined, 'Options')}
+            variant={select<ThumbnailVariant>('Variant', ThumbnailVariant, ThumbnailVariant.squared, 'Options')}
+            imgProps={{ onError }}
+        />
+    );
+};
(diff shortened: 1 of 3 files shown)


packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
```diff
@@ -186,6 +196,7 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
                         src={image}
                         alt={alt}
                         loading={loading}
+                        {...(imgProps || {})}
```
gigeos, line 199: Maybe could you write `{...imgProps}` and set default empty value in params ⬆️

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
```diff
@@ -153,6 +155,7 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
     alt = 'Thumbnail',
     onClick = null,
     focusPoint = DEFAULT_PROPS.focusPoint,
+    imgProps,
```
gigeos, line 158: It would be more consistent to set fallback with empty object here?
`imgProps = {}`

packages/lumx-react/src/components/thumbnail/Thumbnail.stories.tsx
```diff
@@ -49,3 +49,44 @@ export const defaultThumbnail = ({ theme }: { theme: Theme }) => {
...
+                    XXL: Size.xxl,
+                },
+                Size.xxl,
+                'Options',
+            )}
+            theme={theme}
+            isFollowingWindowSize={boolean('Update on window resize', true, 'Options')}
+            resizeDebounceTime={number('Debounce time after resize', 20, undefined, 'Options')}
+            variant={select<ThumbnailVariant>('Variant', ThumbnailVariant, ThumbnailVariant.squared, 'Options')}
+            imgProps={{ onError }}
```
gigeos, line 89: be careful it might not work on IE
https://caniuse.com/#search=img%20onError

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
juanigalan91: Unfortunately, that will trigger a snapshot update on all the components that use `Thumbnail` on the product side, thus making this change a breaking one

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
juanigalan91: Unfortunately, that will trigger a snapshot update on all the components that use `Thumbnail` on the product side, thus making this change a breaking one

packages/lumx-react/src/components/thumbnail/Thumbnail.stories.tsx
juanigalan91: this is just a story, we will do that check on the product side, not here

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
gigeos: Ok :(

gigeos:
Could be merged

packages/lumx-react/src/components/thumbnail/Thumbnail.tsx
```diff
@@ -176,7 +179,14 @@ const Thumbnail: React.FC<ThumbnailProps> = ({
...
             {aspectRatio === AspectRatio.original ? (
-                <img ref={focusImageRef} className={`${CLASSNAME}__image`} src={image} alt={alt} loading={loading} />
+                <img
+                    ref={focusImageRef}
+                    className={`${CLASSNAME}__image`}
+                    src={image}
+                    alt={alt}
+                    loading={loading}
+                    {...(imgProps || {})}
+                />
```
gcornut, line 189: If we want the `imgProps` to not override the props we define in this component, the property spred should be at the top:
```suggestion
                <img
                    {...(imgProps || {})}
                    ref={focusImageRef}
                    className={`${CLASSNAME}__image`}
                    src={image}
                    alt={alt}
                    loading={loading}
                />
```

gcornut: Same thing for the code below

juanigalan91: yep, totally, I will change it

juanigalan91:
@gcornut done!

PR #394, opened 2020-05-12
feat(link-preview): added isThumbnailCrossOriginEnabled prop

Mcourt:
# General summary

Added the `thumbnailProps`  prop to the `LinkPreview` component to be able to control the thumbnail's props.

# Screenshots

# Check list

### DIFF
--- a/packages/lumx-react/src/components/link-preview/LinkPreview.tsx
+++ b/packages/lumx-react/src/components/link-preview/LinkPreview.tsx
@@ -25,6 +25,8 @@ interface LinkPreviewProps extends GenericProps {
     theme?: Theme;
     /** Thumbnail image source */
     thumbnail?: string;
+    /** Active thumbnail cross origin. */
+    isThumbnailCrossOriginEnabled?: boolean;
     /** Link title */
     title?: string;
 }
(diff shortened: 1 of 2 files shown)


packages/lumx-react/src/components/link-preview/LinkPreview.tsx
```diff
@@ -25,6 +25,8 @@ interface LinkPreviewProps extends GenericProps {
     theme?: Theme;
     /** Thumbnail image source */
     thumbnail?: string;
+    /** Active thumbnail cross origin. */
+    isThumbnailCrossOriginEnabled?: boolean;
```
gcornut, line 29: I think it would be better to do:
```tsx
thumbnailProps?: ThumbnailProps;
```
And spread the props in the `<Thumbnail>`
So you could use the LinkPreview like this:
`<LinkPreviewProps thumbnailProps={{isCrossOriginEnabled: true}} />`
What do you think?

Mcourt: @gcornut I thought about it, but that would mean the user could override all of the thumbnail props and therefor alter the `LinkPreview` final design ?
If that's not an issue then yeah totally !

Mcourt: @gcornut Changed according to our PM conversation :)

PR #411, opened 2020-06-09
fix(demo): fix demo site on IE11

gcornut:
# General summary

Minor fixes to make the demo site work again on IE11

# Test scenario

1. run `yarn start`
2. Open http://localhost:4000 in IE11
3. Make sure the website works


gcornut:
Looks like compiling loader-utils to ES5 (for IE11) breaks the loading of page via webpack dynamic import

PR #412, opened 2020-06-10
Refactor popover

pidupuis:
# General summary
Refactor popover and tooltip using popperjs v2

# Screenshots

Peek 17-06-2020 17-44
popover-auto

# Check list


packages/lumx-react/src/components/tooltip/useTooltipOpen.tsx
```diff
@@ -0,0 +1,45 @@
+import { RefObject, useEffect, useState } from 'react';
+
+/**
+ * Hook controlling tooltip visibillity using mouse hover the anchor and delay.
+ * @return whether or not to show the tooltip.
+ */
+export function useTooltipOpen({ delay, anchorRef }: { delay: number; anchorRef: RefObject<HTMLElement> }) {
+    const [timer, setTimer] = useState(0);
```
gcornut, line 8: Use ref instead (should fix bug with persisting tooltip)

packages/lumx-core/src/scss/components/popover/lumapps/_variables.scss
```diff
@@ -0,0 +1,6 @@
+$lumx-popover-arrow-size: 5px !default;
+$lumx-popover-background-color: lumx-color-variant('light', 'N') !default;
+$lumx-popover-color: lumx-color-variant('dark', 'N') !default;
+$lumx-popover-elevation: 3 !default;
+$lumx-popover-font-size: map-get($lumx-typography-font-size, 'caption') !default;
+$lumx-popover-line-height: map-get($lumx-typography-line-height, 'caption') !default;
```
gcornut, line 6: Only the `$lumx-popover-arrow-size` variable is used in the popover style, the others should be removed
```suggestion
```

packages/lumx-angularjs/src/components/dropdown/dropdown_directive.js
```diff
@@ -159,7 +159,7 @@ function DropdownController(
                 .off('scroll', _checkScrollEnd)
                 .insertAfter(_toggleEl);
 
-            if (angular.isDefined(lx.closeOnClick) && !lx.closeOnClick) {
+            if (angular.isDefined(lx.closeOnClickAway) && !lx.closeOnClickAway) {
```
gcornut, line 162: We probably should not touch the angular dropdown

packages/lumx-react/src/components/autocomplete/AutocompleteMultiple.stories.tsx
```diff
@@ -149,6 +149,7 @@ export const simple = ({ theme }: any) => {
             onBlur={onBlur}
             selectedChipRender={renderChip}
             icon={mdiFlag}
+            label="toto"
```
gcornut, line 152: Maybe use a better label?

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Fixed
+
+-   Fix on tooltip that would persist even after the mouse has been moved out of the anchor.
+
+### Changed
+
+-   _[BREAKING]_ `Popover.useComputePosition` hook has been removed.
+-   _[BREAKING]_ `Offset` type does not have `vertical` and `horizontal` keys anymore but has `along` and `away` props.
+-   _[BREAKING]_ `Popover` computes position internally and add popover ref internally. Click away and escape are handled inside the popover. The popover can now have an anchor. `popoverRect`, `isVisible` and `popoverRef` props have been removed. `anchorRef`, `placement`, `offset`, `fitToAnchorWidth`, `isOpen`, `closeOnClickAway`, `closeOnEscape`, `onClose` and `hasArrow` props have been added.
```
gcornut, line 18: I think I would move the `hasArrow` into a separated change like this:

```md
### Added
-      `Popover` now can show an arrow pointing on the anchor using the `hasArrow` prop.
```

gcornut: And actually `closeOnClickAway` and `closeOnEscape` also count as new features for the popover. It's nice to be a little positive and show we don't just break stuff but also add features :D

packages/lumx-react/src/components/dropdown/Dropdown.tsx
```diff
@@ -21,7 +18,7 @@ interface DropdownProps extends GenericProps {
     /** Children of the Dropdown. */
     children: React.ReactNode;
     /** Whether a click anywhere out of the Dropdown would close it. */
-    closeOnClick?: boolean;
+    closeOnClickAway?: boolean;
```
gcornut, line 21: I think we might have misunderstood this prop from the beginning (even before this PR)
If you look at the angular version, the `closeOnClick` should control if the dropdown should close when clicking **inside** it.
So I guess we could have both props here

packages/lumx-react/src/components/autocomplete/AutocompleteMultiple.stories.tsx
pidupuis: You don't say... ^^

Nice catch though

gcornut:
The story "Simple Select With Infinite Scroll" does an infinite loop since it generates list items as long as dropdown is scrolled to the bottom. 
So I guess we really need the automatic height shrink on the dropdown since this could happen in the product with XHR requests bounded to an infinite scroll on a dropdown

gcornut:
Dropdown shrink height now implemented. Still some issues with infinite scroll in the existing select infinite scroll story.

gcornut:
Test OK

PR #415, opened 2020-06-15
fix(chip): fix click propagation on after/before

gcornut:
# General summary

Fix https://lumapps.atlassian.net/browse/LUM-9356

# Test 

1. run `yarn && yarn storybook:react`
2. Go to http://localhost:9000/?path=/story/lumx-components-chip--with-after-and-before
3. Click on the list icon before the chip text (=> you should see "Clicked on chip")
4. Click on the chip text (=> you should see "Clicked on chip")
5. Click on the close icon after the chip text (=> you should see "Clicked on close")

### DIFF
--- a/packages/lumx-react/src/components/chip/Chip.test.tsx
+++ b/packages/lumx-react/src/components/chip/Chip.test.tsx
@@ -1,9 +1,9 @@
-import { ShallowWrapper, shallow } from 'enzyme';
+import { ShallowWrapper, mount, shallow } from 'enzyme';
 import 'jest-enzyme';
-import React from 'react';
+import React, { ReactElement } from 'react';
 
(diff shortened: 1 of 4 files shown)


packages/lumx-react/src/components/chip/Chip.test.tsx
```diff
@@ -127,6 +129,26 @@ describe('<Chip />', () => {
             before.simulate('click', mockClickEvent);
             expect(mockOnAfterClick).not.toHaveBeenCalled();
         });
+
+        it('should not stop propagation when clicking on a "before"or "after" element without an event handler', () => {
```
gigeos, line 133: Missing space :P
```suggestion
        it('should not stop propagation when clicking on a "before" or "after" element without an event handler', () => {
```

PR #417, opened 2020-06-18
Feat - List-item-link: create an <a> if an href prop is provided

radiuso:
# General summary
As the button the `ListItem` component in know able the include a link as wrapper.
This is useful for items in dropdown list like directory-entries.

# Screenshots
Capture d’écran 2020-06-18 à 12 19 34

# Check list


packages/lumx-react/src/components/list/ListItem.tsx
```diff
@@ -144,9 +151,19 @@ const ListItem: React.FC<ListItemProps> = ({
...
             onKeyDown={onKeyDown()}
         >
-            {before && <div className={`${CLASSNAME}__before`}>{before}</div>}
-            <div className={classNames(`${CLASSNAME}__content`)}>{children}</div>
-            {after && <div className={`${CLASSNAME}__after`}>{after}</div>}
+            {useAsLink ? (
+                <a {...linkProps}>
+                    <ListItemContent before={before} after={after} baseClassName={CLASSNAME}>
+                        {children}
+                    </ListItemContent>
```
gcornut, line 158: I don't think we need the `ListItemContent` component. I'd rather repeat the three lines of list item content.
Or we could do 
```jsx
const content = (
    <>
        {before && <div className={`${CLASSNAME}__before`}>{before}</div>}
        className={classNames(`${CLASSNAME}__content`)}>{children}</div>	                            
        {after && <div className={`${CLASSNAME}__after`}>{after}
   </>
);
```
And then:
```jsx
  {useAsLink ? <a...>{content}</a> : <div ...>{content}</div> }
```

radiuso: Yes, why not :)

PR #429, opened 2020-06-27
feat(button): add loading state

radiuso:
# General summary

This is a proposal. 
A button can be in loading state. This case is usefull to handle form loading and prevent the user to click multiple time on the button and create several items
This is the case for https://lumapps.atlassian.net/browse/LUM-10387?atlOrigin=eyJpIjoiMTljOGJkZjAwNmYzNGIyZDgzZjIwMmZiOTFiNjNmZmIiLCJwIjoiaiJ9

# Screenshots
Capture d’écran 2020-06-27 à 16 55 21

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 -   `ListItem` component is now able to use a link as HTML element to wrap content
+-   `Button` component handle isLoading state
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


toto-victoto:
I don't mind adding this seems neat. I usually make the whole modal in loading state in cases like this, or if it's too overkill I would have debounced the apply function.

pidupuis:
Agree with @toto-victoto , inside a dialog, you should use the `isLoading` props of the dialog.

Outside of a modal, the Button can take a `disabled` props and will forward it to the `button` tag, which is widely enough for our cases.

Concerning the animation you offered (which is pretty cool btw), as far as I know, there is no loading state style currently designed for the button. I'll let @matmkian and his team review this.

matmkian:
For now, we have not identify any flow where we need to have a loading button. BUT, I'm going to submit this idea to my fellow designers :-)

PR #435, opened 2020-07-06
fix(list-item): add CSS to fix text-overflow

philippeBchn:
# General summary

L'overflow dans les list item ne marche pas, cf https://lumapps.atlassian.net/browse/LUM-10762

# Screenshots
image

# Check list

-   [X] (if has style) Add `need: review-integration` label
-   [X] (if fix or feature) Add `need: test` label

### DIFF
--- a/packages/lumx-core/src/scss/components/list/lumapps/_mixins.scss
+++ b/packages/lumx-core/src/scss/components/list/lumapps/_mixins.scss
@@ -79,6 +79,8 @@
 
 @mixin lumx-list-item-content() {
     @include lumx-typography('body1');
+    display: flex;
+    min-width: 0;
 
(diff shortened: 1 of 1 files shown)


matmkian:
Why don't you make this change on your markup in the product side? Adding this css rule to the list container can cause problems (display: flex). It depends of what you want to achieve... so you have to do this on your side? Maybe you need helpers to do so without adding css.

philippeBchn:
@matmkian because the DS forces a one line display and try to have a text-overflow but it doesn't work.
Open the Jira, check the problem, to me the bug is inside the DS.

matmkian:
You can put whatever you want in the content of a list item, nothing is forced.

PR #436, opened 2020-07-13
feat(notification-block): implement block in React

gregorycat:
# General summary
Implement Notification Block in React

# Screenshots
image

# Check list

### DIFF
--- a/packages/lumx-core/src/scss/components/notification-block/lumapps/_index.scss
+++ b/packages/lumx-core/src/scss/components/notification-block/lumapps/_index.scss
@@ -3,8 +3,19 @@
    ========================================================================== */
 
 .#{$lumx-base-prefix}-notification-block {
+    $self: &;
+
     display: flex;
     align-items: flex-start;
+    padding: $lumx-spacing-unit * 2 $lumx-spacing-unit * 3;
+
+    &--theme-light:hover {
+        background-color: lumx-color-variant('dark', 'L6');
+    }
+
+    &--theme-dark:hover {
+        background-color: lumx-color-variant('light', 'L6');
+    }
 
     &__before {
         flex-shrink: 0;
@@ -30,7 +41,14 @@
         @include lumx-typography('body1');
 
         display: block;
-        color: lumx-color-variant('dark', 'L2');
+
+        #{$self}--theme-light & {
+            color: lumx-color-variant('dark', 'L2');
+        }
+
+        #{$self}--theme-dark & {
+            color: lumx-color-variant('light', 'L2');
+        }
     }
 
     &__after {
--- a/packages/lumx-react/src/components/notification-block/NotificationBlock.stories.tsx
+++ b/packages/lumx-react/src/components/notification-block/NotificationBlock.stories.tsx
@@ -6,25 +6,26 @@ import { NotificationBlock } from './NotificationBlock';
 
 export default { title: 'LumX components/NotificationBlock' };
 
-export const icon = () => {
+export const icon = ({ theme }: any) => {
     return (
         <div className="demo-grid">
             <NotificationBlock
-                after={<IconButton icon={mdiClose} emphasis={Emphasis.low} />}
-                before={<Icon icon={mdiMessageTextOutline} hasShape />}
+                after={<IconButton icon={mdiClose} emphasis={Emphasis.low} theme={theme} />}
+                before={<Icon icon={mdiMessageTextOutline} hasShape theme={theme} />}
                 title="Title"
                 description="Description message"
                 date="2min ago"
+                theme={theme}
             />
         </div>
     );
 };
 
-export const avatar = () => {
+export const avatar = ({ theme }: any) => {
     return (
         <div className="demo-grid">
             <NotificationBlock
-                after={<IconButton icon={mdiClose} emphasis={Emphasis.low} />}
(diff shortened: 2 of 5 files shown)


packages/lumx-react/src/components/notification-block/NotificationBlock.tsx
```diff
@@ -0,0 +1,92 @@
...
+                className,
+                handleBasicClasses({
+                    prefix: CLASSNAME,
+                    theme,
+                }),
+            )}
+        >
+            {before && <div className={`${CLASSNAME}__before`}>{before}</div>}
+            <div className={`${CLASSNAME}__content`}>
+                <span className={`${CLASSNAME}__title`}>{title}</span>
```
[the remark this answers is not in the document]
radiuso: Could we add stories with a notification without the description etc...

packages/lumx-react/src/components/notification-block/NotificationBlock.tsx
```diff
@@ -0,0 +1,92 @@
...
+                <p className={`${CLASSNAME}__description`}>{description}</p>
+                <span className={`${CLASSNAME}__date`}>{date}</span>
```
radiuso, line 83: @matmkian we have an accessibility warning:`Elements must have sufficient color contrast` when the text color is `dark L2` on the lumapps theme
Should we update the color?

Moreover the component style do not handle border cases such as long text (maybe ellipsis) etc... 
Is it something we should take care here?

And just to be tedious, should we rename the `date` prop? Since we do not format it or use it as a date. So maybe the prop can be more generic like `metadata`?

gregorycat: For long text, we display a multiline paragraph, I'm not sure it's useful to set a max length and an ellipsis. Maybe for the title as we use a `span` tag.

For the date I use a string (like in Angular), but I don't know if we have something in Lumapps to compute the time spent between the notif update and now (To display: `X mins|hours ago`)

packages/lumx-react/src/components/notification-block/NotificationBlock.tsx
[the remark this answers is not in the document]
gregorycat: My bad, the title, description, and date must be mandatory fora notif.

packages/lumx-react/src/components/notification-block/NotificationBlock.tsx
radiuso: I guess we will use the moment function `fromNow` to display what we want. But It's not handled by this component itself

gregorycat: I reproduced the Angular behavior by using a String, it brings a bit more flexibility on what is rendered on the element

matmkian: No update on dark L2 as it's a Material Design color variant -> the contrast checker of chrome is a little bit too hard.
I don't see any cases where we should have ellipsis as we support multiline for every props.
Date seems appropriate as we should display a date but I see your point, I let you decide the best solution.

radiuso: It was an example, but if the responsive is good as is, it's good for me too 😃 
I think we will change the prop name, I let @gregorycat find out what fit the best

Thanks for the answer

packages/lumx-core/src/scss/components/notification-block/lumapps/_index.scss
```diff
@@ -3,8 +3,19 @@
    ========================================================================== */
 
 .#{$lumx-base-prefix}-notification-block {
+    $self: &;
+
     display: flex;
     align-items: flex-start;
+    padding: $lumx-spacing-unit * 2 $lumx-spacing-unit * 3;
```
radiuso, line 10: You can use `$lumx-spacing-unit-big` and `$lumx-spacing-unit-huge` instead

packages/lumx-react/src/components/notification-block/NotificationBlock.stories.tsx
```diff
@@ -6,25 +6,26 @@ import { NotificationBlock } from './NotificationBlock';
 
 export default { title: 'LumX components/NotificationBlock' };
 
-export const icon = () => {
+export const icon = ({ theme }: any) => {
```
radiuso, line 9: Is it required in the DS to have TSX for stories? 
You can use JSX I guess

packages/lumx-core/src/scss/components/notification-block/lumapps/_index.scss
matmkian: No padding please, padding should be implemented according context

gregorycat:
@matmkian The hover style is also implemented according to the context? (Inside Lumapps)

packages/lumx-react/src/components/notification-block/NotificationBlock.stories.tsx
gcornut: I think it's useful to catch minor typing error on component prop values. Plus the only inconvenience is this `any` input for the stories

gregorycat:
Seen with @matmkian. This component has to be included in the DS when some improvement of the notification Block will be designed. For now we can use the List & List Item component to migrate the notification

PR #444, opened 2020-07-29
Minor fixes for tooltip, dropdown and popover

pidupuis:
# General summary

Dropdown: remove limited placement
Tooltip: remove inline-block style
Popover: fix nan maxheight

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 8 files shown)


gcornut:
@juanigalan91 @Flo5k5 With this PR the tooltip won't wrap the anchor in a <div> if it is a Button, a IconButton, an Icon or a React HTML element (any one of them)
This way, if we really want a tooltip around an inline element we just need to make sure the Tooltip wraps a <span> (or other standard React HTML element)
If a custom component is wrapped with a Tooltip it will be wrapped in a div with the `lumx-tooltip-anchor-wrapper`. We did not put any style since we though inline tooltip should be rare

juanigalan91:
Looks great! nicely done

PR #446, opened 2020-07-30
feat(demo-site): update datepicker demo and add prop table

aureldent:
# General summary

Update date-picker demo with real component (not images) and add prop table to it.

# Screenshots

**Before**
Capture d’écran de 2020-07-30 09-53-52

****

**After**
Capture d’écran de 2020-07-30 09-54-44

# Check list

-   [x] (if has textual documentation) Add `need: review-doc` label

### DIFF
--- a/packages/site-demo/content/product/components/date-picker/react/no-past-date-allowed.tsx
+++ b/packages/site-demo/content/product/components/date-picker/react/no-past-date-allowed.tsx
@@ -1,0 +1,26 @@
+import { Moment } from 'moment';
+import React from 'react';
+
+import { DatePicker } from '@lumx/react';
+
+const App = () => {
+    const LOCALE = 'en';
+
+    const now = new Date();
+    const minDate = new Date();
+    minDate.setDate(minDate.getDate() - 1);
+
+    const [datePicked, setDatePicked] = React.useState<Date | Moment | string>(now);
+
+    const handleDatePicked = (picked: Moment) => {
+        setDatePicked(picked);
+    };
+
+    return (
+        <div className="demo-grid">
+            <DatePicker value={datePicked} locale={LOCALE} onChange={handleDatePicked} minDate={minDate} />
+        </div>
+    );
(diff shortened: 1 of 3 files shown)


packages/site-demo/content/product/components/date-picker/react/no-past-date-allowed.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+    const [datePicked, setDatePicked] = React.useState<Date | Moment | string>(now);
+
+    const handleDatePicked = (picked: Moment) => {
+        setDatePicked(picked);
+    };
+
+    return (
+        <div className="demo-grid">
+            <DatePicker value={datePicked} locale={LOCALE} onChange={handleDatePicked} minDate={minDate} />
```
gcornut, line 21: Could this work?
```suggestion
            <DatePicker value={datePicked} locale={LOCALE} onChange={setDatePicked} minDate={minDate} />
```

aureldent: Yes it would work but typescript will complain about type mismatching and the advantage of the handle function is that it made clear the fact that the onChange function return a Moment object and most of the time people will copy paste the snippet and that handle function is where they will put some code :)

gcornut: Or since moment is required in the end, we could simply put:
```tsx

    const now = moment();
    const [datePicked, setDatePicked] = React.useState(now);

    return (
        <div className="demo-grid">
            <DatePicker value={datePicked} locale={LOCALE} onChange={setDatePicked} minDate={now.substract(1, 'days')} />
```

gcornut: But I don't really minde either way

gcornut:
@matmkian can you validate this PR?

PR #456, opened 2020-08-05
fix(tooltip): fix frozen tooltip on button changing its disabled state

gcornut:
# General summary

fix frozen tooltip on button changing its disabled state


pidupuis:
Test OK

PR #459, opened 2020-08-11
fix(mosaic): fix missing CORS props for Thumbnails

Flo5k5:
# General summary

Added `ThumbnailProps` props to the `MosaicElement` interface to be able to forward those props to the `Thumbnail` component inside the `Mosaic`component.

# Screenshots

# Check list

### DIFF
--- a/packages/lumx-react/src/components/mosaic/Mosaic.tsx
+++ b/packages/lumx-react/src/components/mosaic/Mosaic.tsx
@@ -1,14 +1,21 @@
 import React from 'react';
 
-import { AspectRatio, FocusPoint, Theme, Thumbnail } from '@lumx/react';
+import { AspectRatio, CrossOrigin, FocusPoint, Theme, Thumbnail } from '@lumx/react';
 import { COMPONENT_PREFIX } from '@lumx/react/constants';
 import { GenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
 import classNames from 'classnames';
 import take from 'lodash/take';
 
 interface MosaicElement {
-    url: string;
+    /**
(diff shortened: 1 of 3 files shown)


packages/lumx-react/src/components/mosaic/Mosaic.tsx
```diff
@@ -1,14 +1,21 @@
 import React from 'react';
 
-import { AspectRatio, FocusPoint, Theme, Thumbnail } from '@lumx/react';
+import { AspectRatio, CrossOrigin, FocusPoint, Theme, Thumbnail } from '@lumx/react';
 import { COMPONENT_PREFIX } from '@lumx/react/constants';
 import { GenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
 import classNames from 'classnames';
 import take from 'lodash/take';
 
 interface MosaicElement {
```
gcornut, line 9: We could also just replace `MozaicElement` by `ThumbnailProps` and spread every props on the `Thumbnail`, what do you think? 

This pattern is used in other components and it's quite useful.
Although to be very precise to the consumer of the `Mozaic` it would not technically be `ThumbnailProps` but rather `Omit<ThumbnailProps, 'aspectRatio' | 'fillHeight'>` since these props are controlled by the Mozaic component. 
I guess we could define:
`type MozaicElement = Omit<ThumbnailProps, 'aspectRatio' | 'fillHeight'>` :smile:

Flo5k5: Yep seems a good idea to me, it will fix missing FocalPoint issue too.

PR #460, opened 2020-08-12
release v0.21.23

gcornut:
Fix cherry picked and adapted from https://github.com/lumapps/design-system/pull/459


gcornut:
v0.21.23 released. No need to merge as it's a patch on an old version

PR #465, opened 2020-08-24
fix(scss-prefixes-storybook): added postcss on storybook

juanigalan91:
# General summary

Add postcss compilation in storybook


packages/lumx-react/.storybook/postcss.config.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable global-require, import/no-commonjs, import/unambiguous */
+module.exports = {
```
gcornut, line 2: Why not use the `configs/postcss.config.js` directly instead of copying?

juanigalan91: I actually took it from the configurations from the `site-demo` package, I did not see that there was this configuration on the root level. Maybe the one from the site demo should be changed as well?

gcornut: Ah yes I forgot I did a symbolic link from `/configs/postcss.config.js` to `/packages/site-demo/postcss.config.js`.
Indeed it would be simpler just to reference the original config path in both webpack/storybook configurations.

gcornut:
Test OK
2020-09-01-175603_1077x489_scrot

gcornut (approved):

gcornut:
@juanigalan91 I'll let you review my modifications :)

juanigalan91:
@gcornut it is perfect! I can't approve my own PR though so.. you can merge 😅

PR #488, opened 2020-09-07
chore: install chromatic

gcornut:
# General summary

Add chromatic deploy on pull requests

- [x] install chromatic
- [x] install CI auto deploy
- [x] Remove random images in stories (use static image to avoid big diffs)
- [x] Put static id where random uuid are in use (see Switch used in the theme switcher)
- [x] Make pravatar and picsum photos load instantly (preload during webpack build? cache in a SW?)
- [x] Remove the theme switcher from the snapshots ?
- [x] Activate CI only on `need: review-design` label or release branch
- [x] Use secret token


pidupuis (approved):
Sure if you say so

PR #496, opened 2020-09-22
refactor: props harmonization

pidupuis:
# General summary

## Props harmonization

- [x] **CommentBlock** 
  - [x] Create "avataProps" prop (to forward to internal Avatar)
- [x] **Dialog**
  - [x] Remove "onOpen" (user should use "isOpen" instead)
- [x] **ExpansionPanel**
  - [x] Remove "openCallback" & "toggleCallback"
  - [x] Rename "closeCallback" to "onClose"
- [x] **LightBox**
  - [x] Remove "onOpen" prop
  - [x] Add props forwarding
  - [x] Remove "role" prop (cf prop forwarding)
  - [x] Remove "noWrapper" prop
- [x] **LinkPreview**
  - [x] Rename "url" to "link"
  - [x] Add "linkProps" prop forwarding (to forward to internal Link)
- [x] **Mosaic**
  - [x] Change type of "thumbnails" prop from "MosaicElement[]" to "ThumbnailProps[]"
  - [x] Make sure theses forward props to the internal "Thumbnail" components
- [x] **Notification**
  - [x] Rename “handleClick” to “onClick”
  - [x] Rename “actionCallback” to “onActionClick”
- [x] **PostBlock**
  - [x] Remove "HTMLElement" type in props (not valid React type, use ReactNode instead)
  - [x] Add "thumbnailProps" prop (to forward to internal Thumbnail)
  - [x] Remove "thumbnailAspectRatio" prop (use prop forwarding instead)
- [x] **UserBlock**
  - [x] Split "avatar" prop into "avatar: string" and "avatarProps: AvatarProps"
- [x] **Switch**
  - [x] Rename onToggle to onChange
- [x] Homogeneization of onChange, name and value props:
  - [x] Autocomplete
  - [x] AutocompleteMultiple
  - [x] Button
  - [x] IconButton
  - [x] Checkbox
  - [x] DatePicker
  - [x] RadioButton
  - [x] Slider
  - [x] Switch
  - [x] TextField

## Harmonization of Jsdoc definition.

- [x] Fix jsdoc format for each props (no `/* */` anymore, only `/** */`)
- [x] Same sentence format for every prop (`The ... .` for a standard prop and `Whether ... or not.` for a boolean prop)
- [x] Harmonize jsdoc for common props like:
  - [x] `theme`
  - [x] reference
  - [x] `name`
  - [x] `label`
  - [x] `helper`
  - [x] event method
  - [x] `isDisabled`
  - [x] `isSelected`
  - [x] `size`
  - [x] `color`
- [x] Fix and/or add `@see {@link}` when necessary
- [x] Remove extraneous typing like `interface DefaultPropsType extends Partial<IconButtonProps> {}`
- [x] Remove unnecessary (and often not incorrectly copy/paste) component jsdoc
- [x] Sort props:
  - [x] In interface definition
  - [x] in default props definition
  - [x] in props destructuration


packages/lumx-react/src/components/post-block/PostBlock.tsx
```diff
@@ -14,27 +14,27 @@ import { GenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/
...
+    author?: ReactNode;
     /* Meta elements to be transcluded into the component */
-    meta?: HTMLElement | ReactNode;
+    meta?: ReactNode;
     /* Orientation. */
     orientation?: Orientation;
     /* Tags elements to be transcluded into the component */
-    tags?: HTMLElement | ReactNode;
+    tags?: ReactNode;
     /* Content text. Can be either a string, or sanitized html. */
```
gcornut, line 28: Some comments here still use `/*` instead of the JSDoc `/**` format

packages/lumx-react/src/components/select/SelectMultiple.stories.tsx
```diff
@@ -197,7 +197,7 @@ export const chipsCustomSelectMultiple = ({ theme }: any) => {
                 onClick={onClick}
                 theme={theme}
             >
-                -> {choice}
```
gcornut, line 200: :wtf:

packages/site-demo/content/product/components/expansion-panel/react/examples.tsx
```diff
@@ -17,16 +17,22 @@ import {
...
     const [isOpen5, setOpen5] = useState(false);
+    const onClose5 = () => setOpen5(false);
     const [isOpen6, setOpen6] = useState(false);
+    const onClose6 = () => setOpen6(false);
     const stopPropagation = (evt: Event) => evt.stopPropagation();
 
     return (
         <>
-            <ExpansionPanel hasBackground hasHeaderDivider isOpen={isOpen1} toggleCallback={setOpen1}>
+            <ExpansionPanel hasBackground hasHeaderDivider isOpen={isOpen1} onClose={onClose1}>
```
gcornut, line 35: Maybe we could change all the `onClose` callbacks everywhere to be `onClose(open: false): void` so that we could just do `onClose={setOpen}` ?

packages/lumx-react/src/components/expansion-panel/ExpansionPanel.tsx
```diff
@@ -38,14 +31,8 @@ interface ExpansionPanelProps extends GenericProps {
...
     isOpen?: boolean;
 
-    /** The function called on open. */
-    openCallback?: Callback;
-
     /** The function called on close. */
-    closeCallback?: Callback;
-
-    /** The function called on open or close. */
-    toggleCallback?(shouldOpen: boolean): void;
```
gcornut, line 48: I might have forgotten to keep a `onOpen` callback in my props analysis. The expansion panel contains a trigger (the header) that needs to open it.
In that sens it differs from the dropdown, the dialog and the lightbox that need an external trigger
Maybe we could keep these three callback but named `onClode`, `onOpen` and `onToggleOpen` (for convenience)?

packages/lumx-react/src/components/switch/Switch.tsx
```diff
@@ -44,7 +44,7 @@ interface SwitchProps extends GenericProps {
     useCustomColors?: boolean;
```
gcornut, line 45: Should we harmonize the `checked`/`value` props? The `CheckBox` uses `value`, the Switch uses `checked` and the `RadioButton` uses `checked`

Aesthetically I would prefer `value` but I think we probably should use `checked` because it's the HTML standard. The `value` is also used in these elements to have a string representation of the boolean value.
It is especially useful if you name a group of `RadioButton` each having a string value. You basically get a field with enumerated string value rather than multiple boolean values.

Although to make this truly work we probably should change our `onChange` type to `onChange(value: V, name: string, event: SyntheticEvent): void` or `onChange(value: V, name: string, checked: boolean, event: SyntheticEvent): void`?
I'm really not sure on this one, maybe we could debate

gcornut (changes requested):

packages/lumx-react/src/components/post-block/PostBlock.tsx
pidupuis: Yes I saw that but I planned to fix the JSDoc in another PR to let this one more readable.

packages/lumx-react/src/components/select/SelectMultiple.stories.tsx
pidupuis: A warning was triggered during the tests :rofl:

packages/lumx-react/src/components/expansion-panel/ExpansionPanel.tsx
pidupuis: That confirms what I thought. I didn't know how else you expected to expand the panel. I'll fix it.

packages/site-demo/content/product/components/expansion-panel/react/examples.tsx
pidupuis: Good to know!
Won't be necessary since `onToggleOpen` will be finally kept but I'll keep this tip in mind.

packages/lumx-react/src/components/button/ButtonRoot.tsx
```diff
@@ -161,14 +169,20 @@ const ButtonRoot: React.FC<ButtonRootProps> = (props) => {
 
     if (!isEmpty(props.href)) {
         return (
-            <a {...forwardedProps} ref={buttonRef as RefObject<HTMLAnchorElement>} className={buttonClassName}>
+            <a
+                {...forwardedProps}
+                ref={buttonRef as RefObject<HTMLAnchorElement>}
+                className={buttonClassName}
+                aria-disabled={isDisabled}
```
Mcourt, line 176: I'm not sure I see the use case where we would want a disabled link?
If anything, if badly used wouldn't it confuse the user more than it would help him?

For example
```javascript
<Button href="http://google.com"  isDisabled>My profile</Button>
```
focusing this button will cause the screen reader to read
`My profile. Link Disabled` indicating to the user that the link isn't usable. 
But nothing prevents him to press the Enter key and still follow the link.

Maybe there is a use case that I can't think of though ? ^^'

packages/lumx-react/src/components/select/WithSelectContext.tsx
```diff
@@ -180,6 +181,7 @@ const withSelectContext = (
                 targetUuid={targetUuid}
                 anchorRef={anchorRef}
                 isRequired={isRequired}
+                aria-disabled={isDisabled}
```
Mcourt, line 184: I might be wrong, but it seems this can never be actually applied ? 

If I understand the code correctly `WithSelectContext` takes a component as a first `SelectElement` argument.
However, both in the `<Select />` and `<SelectMultiple />` components, the components that are set in the `WithSelectContext` function don't have "aria-disabled" as prop and therefore can never set it on the final component.

I have tried to set the demo component as "isDisabled" but can't fin the aria-disabled anywhere ?

image
image

packages/lumx-react/src/components/slider/Slider.tsx
```diff
@@ -265,6 +265,7 @@ const Slider: React.FC<SliderProps> = ({
                 handleBasicClasses({ prefix: CLASSNAME, theme, hasLabel: Boolean(label) }),
             )}
             onMouseDown={handleMouseDown}
+            aria-disabled={isDisabled}
```
Mcourt, line 268: My screen reader doesn't seem to really use this info :man_shrugging: 
But it can't do any harm either I guess :smile:

packages/lumx-react/src/components/select/WithSelectContext.tsx
gcornut: Yeah I wrongfully assumed the `SelectComponent` would forward the props. Should we put the `aria-disabled` on the `.lumx-select__input-native`?

packages/lumx-react/src/components/button/ButtonRoot.tsx
gcornut: Ok we could either remove the href when `isDisabled` is set or just render a disabled button (but the DOM would alternate between <a> and <button> when toggling the `isDisabled` prop)
What do you think?

pidupuis:
@Mcourt we will treat your feedbacks in an a11y-dedicated PR since there will be a lot of work to do to be a11y compliant and we would like to benefit from the props harmonization asap.

Mcourt:
@pidupuis Deal :)

packages/lumx-react/src/components/button/ButtonRoot.tsx
Mcourt: Yeah you're right, that also might be tad bit confusing. x_x 

For screen reader purpose I would probably, as you said, keep the "a" tag and remove the href attribute.
This way, the screen reader whould either read `My profile. Link` or `My profile. Link disabled`, which correctly describes the link's state. Eslint won't be very fond of it...

I don't now if we want to inforce it through the DS, or leave the responsability to the user to do so though ?
Also, we probably should apply some styling to show that the button is supposed to be disabled ?

packages/lumx-react/src/components/select/WithSelectContext.tsx
Mcourt: Basically you should put it on the focusable element, since it's the element on which the user will land while navigating with his keyboard :) . In that case, the "__wrapper".

Also if you want to go deeper with this component, you could follow the guidelines shown here : https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html#rps_label

I was gonna list them here, but it might be easier and clearer to just read the documentation and see if it's worth applying them now or later. :) 
(This website is pretty great to see how components should behave a11y wise)

gcornut (approved):

PR #500, opened 2020-09-30
chore(github-action): add a release note generator on tag

sebastien-boulle:
A release draft will be generated on github for each tag matching X.Y.Z.

### DIFF
--- a/.github/workflows/release-note.yml
+++ b/.github/workflows/release-note.yml
@@ -1,0 +1,32 @@
+name: Create release note
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


gcornut:
I'm not sure how to compare this with our manual CHANGELOG.md file
It sure is nice to have it automatic but I'm not sure the commit message are enough information for us to properly document evolutions on the Lumx library.
The commit granularity can be both too big or too tiny for a library changelog

What do you think @matmkian @pidupuis ?

pidupuis:
I agree that we need to continue to manually redact a CHANGELOG to only list useful informations.

PR #517, opened 2020-10-14
chore: upgrade to mdi 5.8.55 and handle backward compatibility LUM-12974

pidupuis:
# General summary

Upgrade `mdi` to 5.6.55 and handle backward compatibility.
`mdi` has been upgraded to v5x first in [v0.26.0](https://github.com/lumapps/design-system/releases/tag/v0.26.0) but then downgraded back to v4x in [v0.26.1](https://github.com/lumapps/design-system/releases/tag/v0.26.1) for stability purpose.

To remember when upgrading into LumApps product:

mdiFacebookBox (to remove from icon picker)
mdiGithubFace (to remove from icon picker)
mdiLinkedinBox (to remove from icon picker)
mdiTwitterBox (to remove from icon picker)
mdiGooglePlusBox (don't exists anymore)

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Added component `SlideshowControls` to control a slideshow from the outside
 -   Add `linkAs` prop on `Button` and `IconButton` to customize the link component (can be used to inject the `Link` component from `react-router`).
+-   Make it possible to override `tabIndex` and `role` props of `ListItem.linkProps`.
+
+### Fixed
+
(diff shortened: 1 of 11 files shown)


gcornut:
We could automatically detect deleted icons and modified icons using the icon identifier before the update and after the update
before: https://raw.githubusercontent.com/Templarian/MaterialDesign-SVG/v4.2.95/meta.json
after: https://raw.githubusercontent.com/Templarian/MaterialDesign-SVG/v5.6.55/meta.json

Lumapps Web also seem to use icon aliases which is trickier to handle in @lumx/icons because two icons can use the same alias...

gcornut:
Need to test in the product

gcornut (approved):

PR #540, opened 2020-11-12
refactor(tabs): separate panels from tabs and improve a11y LUM-11309

pidupuis:
# General summary
-   _[BREAKING]_ Removed `Tabs` component replaced by the `TabProvider` and `TabList` components.
-   _[BREAKING]_ `Tab` component doesn't wrap the tab content anymore (use `TabPanel` for that). Implements the [WAI-ARIA `tab` role](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html#rps_label).
-   Added `TabProvider` component wrapping the new `TabList` and `TabPanel` and handling state (controlled or uncontrolled).
-   Added `TabPanel` component that wraps the content of a tab (previously wrapped in the `Tab` component). Implements the [WAI-ARIA `tabpanel` role](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html#rps_label).
-   Added `TabList` component that wraps `Tab` components. Implements the [WAI-ARIA `tablist` role](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html#rps_label).


packages/lumx-react/src/components/tabs/TabList.tsx
```diff
@@ -80,12 +88,10 @@ const Tabs: React.FC<TabsProps> = ({
             })}
         >
             <div className={`${CLASSNAME}__links`}>{tabs}</div>
```
Mcourt, line 90: A few things are missing a11y-wise to be perfect here.

* The div wrapping the tabs should have the `tablist` role (that's easy :p)
* In a perfect world, it should also have either a `aria-label` or `aria-labelledby` attribute to explain the purpose of the tabs (ex : Settings). I understand we might not want to force users to set them, but it would be nice to be able to when using the component. 
So either we destructure the props "aria-label" or "aria-labelledby" so we can set them on the wrapper (whether at least one of them should be required or not is up to debate), or maybe we could add a "wrapperProps" that is spread here ? 

```suggestion
            <div className={`${CLASSNAME}__links`} role="tablist" aria-label={ariaLabel}>{tabs}</div>
```

Or 

```suggestion
            <div className={`${CLASSNAME}__links`} role="tablist" {...wrapperProps}>{tabs}</div>
```

packages/lumx-react/src/components/tabs/Tab.tsx
```diff
@@ -61,23 +63,26 @@ const Tab: React.FC<TabProps> = ({
...
         }
 
-        onTabClick?.({ event, index });
+        onChange?.(tabKey as string);
     };
 
     return (
-        <a
+        <button
+            aria-controls={tabKey}
```
Mcourt, line 71: To be a11y complient, the tabs really should have the `tab` role.

```suggestion
            role="tab"
            aria-controls={tabKey}
```

that way assistive technologies can make the link between the `tablist` parent and the actual tabs.

Also `aria-controls` has to be the id of the `tabpanel` element linked to this tab. (the content). Therefore setting it without having set a corresponding id to it's content might be confusing for the user :/ (And also, it might break our Lighthouse score... which I'm pretty sure someone will check at some point).
I'm not sure how we could make it easier for the user to make the link between the two... 

Also, regarding the tab panels themselves, they should have the following attributes to be complient

```
  role="tabpanel"
  id="someUniqueId"
  aria-labelledBy="theTabId"
```

Maybe we could add a TabPanel component to manage this ? I don't know... Just a though

Mcourt (commented):
Added some comments about roles and aria that should be set.
However, @pidupuis and I had a private conversation about the missing keyboard behavior. We knew we wouldn't have time to add them but it raises the question on if we should still add the `role` attributes if we don't have the corresponding behavior. :thinking: 

Ex: when adding a `role=tabList` with `role=tab`, a user using assistive technology expects
* To only be able to focus the active tab.
* To use left/right to change tab
* To press space or enter to activate the tab
* The activated tab content should have tabIndex="0" to be in the tab order.

It might not be a problem for tabs though... I'm not really sure to be honest :(

packages/lumx-react/src/components/tabs/Tab.tsx
pidupuis: I don't know we could enforce a11y on tab panels since they are not DS component.
About `aria-controls`, I am well aware and that's why I had written the following as possible improvements:
Capture d’écran de 2020-11-17 09-10-56

We should discuss about the tab panels. Maybe we should create an empty wrapper `TabPanel` including a11y attributes :man_shrugging: What do you think ?

Thanks for the `role`s btw, I wasn't reading correctly the table from https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html (weird format imo)

packages/lumx-react/src/components/tabs/TabList.tsx
pidupuis: The first suggestion seems good to me. I don't know if we should enforce it but having it as a props can help, even more if the documentation and demo include it.

Mcourt: Returning on this, it might actually not be a problem to add the role + label on the parent element (the one that has the forwardedProps). As you wish :)

packages/lumx-react/src/components/tabs/TabList.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+        return cloneElement(tab, {
+            id: tab.props.id as string,
+            isActive,
+            /** Tabs should only have one focusable tab at a time. It should be the active tab. */
+            tabIndex: isActive ? 0 : -1,
+            onChange,
+            onFocus,
+        });
+    });
```
Mcourt, line 92: I realised we forgot about blocking selection when a tab is disabled :grimacing: 

Could we maybe do something like this ?
```suggestion
    const tabs: ReactElement[] = Children.map(children as ReactElement[], (tab: ReactElement, index: number) => {
        const { id: tabId, isDisabled } = tab.props;
        /** If no activeTabId, the first tab is considered active. */
        const isActive = activeTabId === tabId || (!activeTabId && index === 0);

        return cloneElement(tab, {
            id: tabId as string,
            isActive,
            /** Tabs should only have one focusable tab at a time. It should be the active tab. */
            tabIndex: isActive ? 0 : -1,
            onChange: !isDisabled && onChange,
            onFocus: !isDisabled && onFocus,
        });
    });
```

packages/lumx-react/src/components/tabs/TabList.test.tsx
```diff
@@ -96,36 +97,21 @@ describe(`<${Tabs.displayName}>`, () => {
...
-    // 4. Test conditions (i.e. things that display or not in the UI based on props).
-    describe('Conditions', () => {
-        it('should fail when no `Tab` children is given', () => {
-            expect(() => {
-                // We know that children must be given to <Tabs>, but for the test, ignore it.
-                setup({ children: null });
-            }).toThrowErrorMatchingSnapshot();
+            expect(onChange).toHaveBeenCalledWith('tab-1');
         });
     });
```
Mcourt, line 113: Following the "isDisabled" modification, here's a new test to check that it works properly. :) 

```suggestion

        it('should not trigger `onChange` when a disabled child tab is clicked', () => {
            const tabs = [
                <Tab id="tab-0" label="Tab 0" aria-controls="tab-panel-0" />,
                <Tab id="tab-1" label="Tab 1" aria-controls="tab-panel-1" isDisabled />,
            ];
            const { wrapper } = setup({ onChange, children: tabs }, false);
            const secondTab = wrapper.find('Tab[id="tab-1"]');

            secondTab.simulate('click');
            expect(onChange).not.toHaveBeenCalled();
        });
    });
```

Mcourt:
@pidupuis I have just commited a few changes and improvements :grimacing: 

* removed "onChange" and "onFocus" on disabled tabs https://github.com/lumapps/design-system/pull/540/files#diff-464ecc0cbb2849ad0e5e25b80e2e17cb2a6c4bdcddeaddc55cea3d5a1d888b6cR90-R91
* Added some tests for disabled tabs https://github.com/lumapps/design-system/pull/540/files#diff-bfad564c7d8992dab361a4c6c17c4e353dd58e736941ca3bc8a3a6813eb0ed09R114-R124
* Changed my mind on the role and aria-label of the TabList :grimacing: My bad, I tried it out with a screen reader and it works better like this https://github.com/lumapps/design-system/pull/540/files#diff-464ecc0cbb2849ad0e5e25b80e2e17cb2a6c4bdcddeaddc55cea3d5a1d888b6cR103-R105
* Updated the snaps for this last change :point_up:

packages/lumx-react/src/hooks/useRovingTabIndex.tsx
```diff
@@ -0,0 +1,84 @@
...
+        const handleKeyDown = (index: number) => (evt: KeyboardEvent) => {
+            let newTabFocus = index;
+            if (!(evt.key === 'ArrowRight' || evt.key === 'ArrowLeft')) {
+                return;
+            }
+
+            let newElement: HTMLElement | undefined;
+            if (evt.key === 'ArrowRight') {
+                // Move right
+                while (!newElement || get(newElement, 'attributes.disabled')) {
```
Mcourt, line 29: Was the while really necessary ? I'm not sure I see the gain from using it when all our targeted elements are already in a nodeList :/ 
Using a "while" seems like a good way of introducing an infinite loop IMO.

Also, disabled elements need to be focused when using the roving tabIndex technique, at least for tabs they should.
https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls
```
For the following composite widget elements, keep them focusable when disabled:
* Options in a Listbox
* Menu items in a Menu or menu bar
* Tab elements in a set of Tabs
* Tree items in a Tree View
```

packages/lumx-react/src/components/tabs/TabProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * It does not produce any markup so you can wrap it around any React elements and then split the TabList and TabPanel
+ * components in the react tree.
+ *
+ * @param  props React component props.
+ * @return React element.
+ */
+export const TabProvider: TabProvider = (props) => {
+    const { children, shouldActivateOnFocus, activeTabIndex, isLazy, onChange } = props;
+    const [state] = useState<TabContextState>(
```
Mcourt, line 71: Having that much logic in a useState seems really odd to me...
I might be wrong, but couldn't we have used a useReducer instead ?

packages/lumx-core/src/scss/components/tabs/lumapps/_index.scss
```diff
@@ -119,3 +119,21 @@
...
     }
 }
+
+/* Tab panel
+   ========================================================================== */
+
+.#{$lumx-base-prefix}-tab-panel {
+    display: none;
+    outline: none;
+
```
Mcourt, line 129: I get where you are coming from, but this is an antipattern accessibility whise...
Having a focusable element without an outline is probably as confusing for the user as not having the element focusable.

If we don't like the "focused" style, it should be improved not removed.

packages/lumx-react/src/components/tabs/Tab.tsx
```diff
@@ -39,45 +36,80 @@ const CLASSNAME = `${CSS_PREFIX}-tabs__link`;
...
+            id={tabId}
             className={classNames(className, handleBasicClasses({ prefix: CLASSNAME, isActive, isDisabled }))}
-            tabIndex={isDisabled ? -1 : 0}
-            onClick={handleTabClick}
+            onClick={handleClick}
             onKeyPress={handleKeyPress}
+            onFocus={handleFocus}
+            role="tab"
+            tabIndex={isActive ? 0 : tabIndex}
+            disabled={isDisabled}
```
Mcourt, line 102: As noted in another comment (https://github.com/lumapps/design-system/pull/540/files#r527516518), disabled tabs need to still be focusable (but not clickable).
This is why only the aria-disabled was used + click was handled to not trigger when tab was disabled.

Adding the "disabled" attribute completely blocks the focus from being set on the tab...

packages/lumx-core/src/scss/components/tabs/lumapps/_index.scss
gcornut: It's the anti pattern used on the whole LumX design system xD
We remove the focus tab on click focus and add it back on tab focus

Mcourt: Have you tried it ? It doesn't work on my end, there is no outline ring on tab focus. :( 
It comes back when I remove this line though.

packages/lumx-react/src/hooks/useRovingTabIndex.tsx
gcornut: removing

packages/lumx-react/src/components/tabs/TabProvider.tsx
gcornut: I probably would have started with a useReducer if I knew what I was getting into with this component ^^
But I am indeed thinking how to adapt into a reducer. It's a shame we don't have slice in this repo :/

packages/lumx-react/src/components/tabs/Tab.tsx
gcornut: removing

packages/lumx-core/src/scss/components/tabs/lumapps/_index.scss
gcornut: It was the style I used that was bad :/ 
It should work now

Mcourt: Nice! Sorry about the initial comment, I spoke too soon and misread!

gcornut:
@matmkian Could you review the SCSS and test the new tab component?

gcornut (approved):

PR #544, opened 2020-11-16
feat/Truncate - Add the util

radiuso:
# General summary

Add an util to be able to truncate a text like
"juani is amazing" -> "juani is..."

It can be used for chips and any other component

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Added `name` and `value` props to `Switch`, `CheckBox` and `RadioButton` components.
 -   Added `name` prop to `Autocomplete`, `DatePicker`, `TextField`, `Button`, `IconButton` and `Slider` components.
 -   Expose component default props in React `Component.defaultProps`
--   Added util `truncate` that allow a text to be truncate at the `end`, `start`, or `middle`.
+-   Added `Text` component that allow a text to be truncate at the `end`, `start`, or `middle`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Added `name` and `value` props to `Switch`, `CheckBox` and `RadioButton` components.
 -   Added `name` prop to `Autocomplete`, `DatePicker`, `TextField`, `Button`, `IconButton` and `Slider` components.
 -   Expose component default props in React `Component.defaultProps`
+-   Added util `truncate` that allow a text to be truncate at the `end`, `start`, or `middle`.
 
 ### Changed
 
--- a/packages/lumx-react/src/components/text/Text.tsx
+++ b/packages/lumx-react/src/components/text/Text.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+
+import { COMPONENT_PREFIX } from '@lumx/react/constants';
+import { EllipsisType, GenericProps, truncate } from '@lumx/react/utils';
+
+/**
+ * Defines the props of the component.
+ */
+interface TextProps extends GenericProps {
+    /** The text to display */
+    text: string;
+    /** The max number of characteres to display */
+    maxChars?: number;
+    /** Works with maxChars. The ellipsis method to use */
+    ellipsisType?: EllipsisType;
+}
+
+/**
+ * The display name of the component.
+ */
+const COMPONENT_NAME = `${COMPONENT_PREFIX}Text`;
+
+/**
+ * The default value of props.
+ */
+const DEFAULT_PROPS: Partial<TextProps> = {
+    maxChars: -1,
+    ellipsisType: EllipsisType.END,
+};
+
+const Text: React.FC<TextProps> = ({ text, ellipsisType, maxChars, ...forwardedProps }) => {
+    // const shouldTruncate = ellipsisType && maxChars && maxChars > 0 && text.length > maxChars;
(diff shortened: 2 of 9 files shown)


toto-victoto:
Isn't css already able to do this by itself? Do we really want to handle this manualy?

gcornut:
Looks good to me. However we currently only export `packages/lumx-react/src/index.ts` in the lib `@lumx/react`.
To have a `@lumx/react/utils` module we would need to modify the webpack config and add an entry to it.

@matmkian Do you think we could use a `<Text>` component that could apply this truncate, the dark/light theme, the typography?

matmkian:
@gcornut Good question -> what are "competitors" doing regarding that?

packages/lumx-react/src/components/text/Text.tsx
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * The default value of props.
+ */
+const DEFAULT_PROPS: Partial<TextProps> = {
+    maxChars: -1,
+    ellipsisType: EllipsisType.END,
+};
+
+const Text: React.FC<TextProps> = ({ text, ellipsisType, maxChars, ...forwardedProps }) => {
+    // const shouldTruncate = ellipsisType && maxChars && maxChars > 0 && text.length > maxChars;
```
pidupuis, line 32: Dead code

packages/lumx-react/src/components/text/Text.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const a11yProps: any = {};
+    const resultingText = truncate(text, { max: maxChars as number, ellipsisType });
+
+    if (resultingText !== text) {
+        a11yProps.ariaLabel = text;
```
pidupuis, line 37: I wouldn't put the `aria-label` here. At least not automatically.

I think this is something that should be handled by the consumer of the component. Here you are saying that sighted users should see a truncated version of the text but visually impaired users should have the entire text. In some cases, this wouldn't be applicable, for example if you are truncated a huge amount of text.

If you look at the search page result of W3 site: https://www.w3.org/WAI/search/?q=accessible
You'll see multiple truncated result without any particular aria-label. Sighted and visually impaired users would have the same experience.

Concerning small texts like title and link, it is common to truncate them but still offer to the user the access to the complete information. In this case, there will be an aria-label for visually impaired users but for sighted users, you would have a tooltip.
This makes me think that the `aria-label` shouldn't be set by the `Text` component. It should be handled by the developer when using `Text` by surrounding it by a `Tooltip` which handles the `aria-label` itself.

What do you think?

packages/lumx-react/src/utils/truncate.ts
```diff
@@ -0,0 +1,41 @@
...
+    MIDDLE = 'middle',
+}
+
+export interface TruncateParams {
+    max: number;
+    ellipsisType?: EllipsisType;
+    separator?: string;
+}
+
+export const ELLIPSIS_CHARS = '...';
```
pidupuis, line 13:
```suggestion
export const ELLIPSIS_CHARS = '…';
```
You should use the dedicated ellipsis character instead.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Added `name` and `value` props to `Switch`, `CheckBox` and `RadioButton` components.
 -   Added `name` prop to `Autocomplete`, `DatePicker`, `TextField`, `Button`, `IconButton` and `Slider` components.
 -   Expose component default props in React `Component.defaultProps`
+-   Added `Text` component that allow a text to be truncate at the `end`, `start`, or `middle`.
```
pidupuis, line 20:
```suggestion
-   Added `Text` component that allows a text to be truncated at the `end`, `start`, or `middle`.
```

packages/lumx-react/src/components/text/Text.tsx
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Defines the props of the component.
+ */
+interface TextProps extends GenericProps {
+    /** The text to display */
+    text: string;
+    /** The max number of characteres to display */
+    maxChars?: number;
+    /** Works with maxChars. The ellipsis method to use */
+    ellipsisType?: EllipsisType;
```
pidupuis, line 15: These props make sense only in the context of the truncate util so I would have create a unique prop:

```suggestion
    /** Defined how the text should be truncated. */
    truncate?: TruncateParams;
```

Because from my point of view, either you have a `Text` component with `text` and `truncate` props, or you have a `Truncate` component with `max`, `ellipsis` and `separator` props.

packages/lumx-react/src/components/text/Text.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * The display name of the component.
+ */
+const COMPONENT_NAME = `${COMPONENT_PREFIX}Text`;
+
+/**
+ * The default value of props.
+ */
+const DEFAULT_PROPS: Partial<TextProps> = {
+    maxChars: -1,
+    ellipsisType: EllipsisType.END,
```
pidupuis, line 28: Generally, when we have optional props with _falsy_ values, we prefer not to set them.

packages/lumx-react/src/utils/truncate.ts
pidupuis: By the way, I don't think this constant is necessary and I don't think this is necessary to export it.

gcornut:
Thank you @radiuso for this PR! Like we discussed, we are not yet very sure how we want to introduce the text truncate and/or the Text component. 
We'll explore and prototype all of this in the product and once the API is settled we'll add it to this repo.
For now we will close the PR. Thanks again!

PR #551, opened 2020-11-26
refactor(date-picker): value prop should be a date LUM-13240

pidupuis:
# General summary

-   _[BREAKING]_ `DatePicker` and `DatePickerField` `value` prop can no longer be a `Moment` object. This will allow us to remove `moment` dep in the future without generating breaking change later. The `value` prop can no longer be a `string` to avoid handling incompatible values. `value` can only be a `Date` (or `undefined` for non selectable values).

# Check list


gcornut (approved):

gcornut:
Test OK

PR #560, opened 2020-12-02
Fix popover bugs

pidupuis:
# General summary

- [x] Update placement when children change: https://lumapps.atlassian.net/browse/LUM-11530

Before
before
After
after

- [x] Scrolling inside popover should not modify its position so drastically: https://lumapps.atlassian.net/browse/LUM-13372 based on https://lumapps.atlassian.net/browse/LUM-13128. This is due to `applyMaxHeight` modifier when an `overflow: auto` is added to the popover like for notification center. However, without `overflow: auto`, the entire page is scrolling like we can see in `withScrollingPopoverContent` story. The `fitWithinViewportHeight` prop does not work well.

- [x] Click on scrollbar should not close the popover. This is due to the fact that the `ClickAwayProvider` does not consider the scrollbar to be part of the popover.

- [x] Fix arrow placement on placement variant `start` & `end` (only worked with `top`, `bottom`, `right` and `left`)
Peek 02-12-2020 16-36

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Fixed ClickAwayProvider initialization error using deferred initialization.
 -   Fixed `Popover` placement on children update.
+-   Fixed `Popover` and `Dropdown` scrollbar behavior.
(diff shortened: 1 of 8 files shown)


gigeos (approved):

gcornut (approved):

gcornut:
Test OK

PR #586, opened 2020-12-29
Fix/LUM-13688 - Chips color contrast

radiuso:
# General summary

The color contrast on Chip wasn't satisfying a11y rules

# Screenshots

Screenshot 2020-12-29 at 17 06 51

# Check list

### DIFF
--- a/packages/lumx-core/src/scss/core/state/lumapps/_mixins.scss
+++ b/packages/lumx-core/src/scss/core/state/lumapps/_mixins.scss
@@ -31,7 +31,7 @@
 @mixin lumx-state-medium($state, $color, $has-focus-inset: false) {
     @if $state == lumx-base-const('state', 'DEFAULT') {
         background-color: lumx-color-variant($color, 'L5');
-        color: lumx-color-variant($color, 'N');
+        color: lumx-color-variant($color, 'D3');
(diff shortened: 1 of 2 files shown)


gcornut:
LGTM code-wise. We'll need @matmkian validation on the styling/integration aspect

@radiuso Is this a11y issue appearing in both material and lumapps themes? We currently only use the material theme in the product and the lumapps theme color palette is likely to be removed

matmkian:
D3 variant doesn't exists.... Our state rules are applied everywhere, so we have to take the accessibility issue into consideration at the color palette level ans the global states level with @FlorentDinet

FlorentDinet:
Where the colors in the sceenshot come from ?

I've this on the DS [website material theme ](https://design.lumapps.com/product/components/chip/)
Capture d’écran 2021-01-04 à 12 18 01

radiuso:
@FlorentDinet this come from the storybook.
As you see even on the DS site the contrast ratio is not enough. 
We expect a ratio of 4.5:1.
We can change it in the color palette if it is more relevant.

@gcornut I have tested mostly on the lumapps theme

matmkian:
LumApps theme is not used and we cannot change the MaterialDesign color palette quickly. Let's wait to change from Material Design to LumApps theme and set the appropriate color palette.

PR #616, opened 2021-01-28
fix(tooltip): fix closing on covered anchor

gcornut:
`mouseenter`/`mouseleave` event on the tooltip anchor element do not trigger when an element appear between the anchor element and the mouse

The solution implemented here is to put a `mouseover` event listener on the `document` and check that the element currently being hovered is one of the anchors of the registered tooltips to open and close them.


pidupuis (approved):

gcornut:
Test OK

PR #629, opened 2021-02-16
Feat/Flag component

radiuso:
# General summary

https://www.figma.com/file/oBsCET5MB3dQMDkBAc214t/DS%3A-material-theme?node-id=0%3A1

Create the `Flag` component.
This component should be use in the R34 in the search context.

# Screenshots

Screenshot 2021-02-16 at 17 26 39

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,13 +12,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Fixed `TabProvider` & `ProgressTrackerProvider` not calling `onChange` when the index is not provided via the props.
 -   Fixed `Link` default style when in button mode (padding + text align).
 
+### Added
+
+-   New component in React: `Flag`
+
 ## [1.0.8][] - 2021-02-08
 
 ### Added
 
 -   Added `light` theme by default on Skeleton components.
 -   Added `badge` prop on `Thumbnail` component.
--   New component in React: `Flag`
 
 ## [1.0.7][] - 2021-02-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Added `light` theme by default on Skeleton components.
 -   Added `badge` prop on `Thumbnail` component.
--   New `@lumx/core` CSS variable customization.
-    -   New lumapps theme applied by default
-    -   Old Material theme applicable using variables in `css/material.css`
 -   New component in React: `Flag`
 
 ## [1.0.7][] - 2021-02-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Added `light` theme by default on Skeleton components.
 -   Added `badge` prop on `Thumbnail` component.
+-   New `@lumx/core` CSS variable customization.
+    -   New lumapps theme applied by default
+    -   Old Material theme applicable using variables in `css/material.css`
+-   New component in React: `Flag`
 
 ## [1.0.7][] - 2021-02-03
 
--- a/packages/lumx-react/src/components/flag/Flag.test.tsx
+++ b/packages/lumx-react/src/components/flag/Flag.test.tsx
@@ -4,9 +4,11 @@ import React, { ReactElement } from 'react';
 
 import { ColorPalette, Theme } from '@lumx/react';
 import { mdiAbTesting } from '@lumx/icons';
-import { Wrapper } from '@lumx/react/testing/utils';
+import { itShouldRenderStories, commonTestsSuite, Wrapper } from '@lumx/react/testing/utils';
 import { getBasicClass } from '@lumx/react/utils';
+
 import { Flag, FlagProps, CLASSNAME } from './Flag';
+import * as stories from './Flag.stories';
 
 /**
  * Mounts the component and returns common DOM elements / data needed in multiple tests further down.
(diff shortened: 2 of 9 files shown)


packages/lumx-react/src/components/flag/Flag.tsx
```diff
@@ -0,0 +1,42 @@
...
+    theme: Theme.light,
+};
+
+/**
+ * Flag component.
+ *
+ * @param  props Component props.
+ * @return React element.
+ */
+export const Flag: React.FC<FlagProps> = ({ label, icon, color, className, theme }) => {
```
pidupuis, line 31: Could you add a `forwardRef`? Every component should be able to forward a ref to be usable in some context which requires a ref like using this component as an anchor for a popover or a dropzone for a drag and drop feature, and so on...

You can look at the Badge component for example, except that in your case the type of your ref would be `HTMLSpanElement`.

packages/lumx-react/src/components/flag/Flag.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const flagColor = color || (theme === Theme.light ? ColorPalette.dark : ColorPalette.light);
+
+    return (
+        <span className={classNames(className, handleBasicClasses({ prefix: CLASSNAME, color: flagColor }))}>
```
pidupuis, line 35: In addition to forward a ref on this component, could you also spread a `forwardedProp` variable? It would allow us to pass any compatible props that could be useful in some contect (`id`, `data-id`, and so on)

packages/lumx-react/src/components/flag/Flag.tsx
```diff
@@ -0,0 +1,42 @@
...
+            {icon && <Icon icon={icon} color={color} colorVariant={ColorVariant.D2} className={`${CLASSNAME}__icon`} />}
+            {label}
+        </span>
+    );
+};
+Flag.displayName = COMPONENT_NAME;
+Flag.defaultProps = DEFAULT_PROPS;
```
pidupuis, line 42: Could you add the following?
```
Flag.className = CLASSNAME;
```

packages/lumx-react/src/components/flag/Flag.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return {
+        props,
+        iconEl: wrapper.find('Icon'),
+        wrapper,
+    };
+};
+
+describe('<Flag />', () => {
+    // 1. Test render via snapshot (default state of component).
+    describe('Snapshot', () => {
```
pidupuis, line 31: Could you use `itShouldRenderStories` instead? This method will generate snapshot based on your stories which cover more cases that your current test.

Also, could you use `commonTestsSuite`? This method tests that your component can forward props and className properly.

Here again, you can look at the Badge component for inspiration.

packages/lumx-react/src/components/flag/Flag.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export interface FlagProps extends GenericProps {
+    /** Theme adapting the component to light or dark background. */
+    theme?: Theme;
+    /** The color of the component */
+    color?: ColorPalette;
+    /** icon */
+    icon?: string;
+    /** label of the flag */
+    label: string;
```
pidupuis, line 16: Could you please standardize the jsdoc? This means dropping the `The` formalism, put an uppercase at the beginning of the sentence and a dot at the end.
This is less important inside the product codebase because the product jsdoc is poorly used. However, the jsdoc of the DS is used inside the public documentation (in the table props).

pidupuis (changes requested):
Thanks a lot for this contribution @radiuso!

I've asked for some slightly modifications though, which are only to ensure that our components are as uniform as possible.

gcornut (changes requested):
:+1: On all the feedback from pidup
Could you also rebase the PR onto `release/v1` ? 
The master branch has a major modification that we are not ready to release yet. So we will release new fix and features from `release/v1` and then we'll merge it back in master

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Added `light` theme by default on Skeleton components.
 -   Added `badge` prop on `Thumbnail` component.
+-   New component in React: `Flag`
```
gcornut, line 16: This must be in the unreleased section, not here

pidupuis, line 16: Should be put in `Unreleased` section under a new `Added` subsection.

packages/lumx-react/src/components/flag/Flag.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    return {
+        props,
+        iconEl: wrapper.find('Icon'),
+        flag: wrapper.find('span'),
+        wrapper,
+    };
+};
+
+describe('<Flag />', () => {
```
pidupuis, line 32:
```suggestion
describe(`<${Flag.displayName}>`, () => {
```

gcornut (changes requested):

packages/lumx-react/src/components/flag/Flag.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const CLASSNAME = getRootClassName(COMPONENT_NAME);
+const DEFAULT_PROPS: Partial<FlagProps> = {
+    theme: Theme.light,
+};
+
+/**
+ * Flag component.
+ *
+ * @param  props Component props.
+ * @return React element.
```
pidupuis, line 28:
```suggestion
 * @param  ref   Component ref.
 * @return React element.
```

pidupuis (changes requested):

PR #630, opened 2021-02-18
fix(tabs): remove default props preventing onChange from being called

gcornut:
The default props for TabProvider and ProgressTrackerProvider was preventing the onChange to be called when the parent component doesn't update the `activeTabIndex` prop.

This PR removes the active index default prop (the reducer initial state already initializes the index correctly).

Test scenario:
- Check "Tabs > Not controlled" story.
- Clicking on any tabs should log it's index in the console.
- Repeat the test on "Progress tracker > Not controlled" story.


pidupuis (approved):

pidupuis:
Test OK

PR #639, opened 2021-02-25
fix(avatar): use default aspect ratio

pidupuis:
# General summary

-   `Thumbnail` default aspect ratio for `Avatar` is now `square`. This is not a breaking change since the style has been modified to result in the same render as before. This modification prevents rectangular avatars to have a broken ratio.

# Check list


packages/lumx-react/src/components/avatar/Avatar.tsx
```diff
@@ -47,6 +51,9 @@ const CLASSNAME = getRootClassName(COMPONENT_NAME);
 const DEFAULT_PROPS: Partial<AvatarProps> = {
     size: Size.m,
     theme: Theme.light,
+    thumbnailProps: {
```
matmkian, line 54: Shouldn't we add the AspectRatio prop on the thumbnailProps omit list as we want to force square aspect ratio (avatar is always a circle)?

matmkian (changes requested):

matmkian (approved):

gcornut (approved):

PR #673, opened 2021-06-28
fix(click-away): fix nested click away using React ref

gcornut:
# General summary

Fixing nested click away not working properly. Replacing inner state by refs


Chocazimut:
Tests : 

✅ : click away from select 
✅ : click away from date picker

pidupuis (approved):

PR #674, opened 2021-06-28
fix(select): fix focus state not updating properly

gcornut:
# General summary

The focus state of the select was not activating using keyboard navigation and would not deactivate when clinking outside the select. 
I'm refactoring the hook used to listen to this state using a correct element reference.

**Tests:**
Example page: https://5fbfb1d508c0520021560f10-tyxrbjhdfg.chromatic.com/?path=/story/lumx-components-dialog-dialog--dialog-with-focusable-elements
1. Check the select get the focus correctly using Tab key nav or directly clicking on it
2. Check the select looses the focus when clicking outside or navigating to another focusable element

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   Fix Select/Dropdown inside Dialog not closing when clicking outside.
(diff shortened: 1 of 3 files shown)


Chocazimut:
tested : 
✅ focus state propoerly updated when click outside
✅ focus state propoerly updated when select an option

pidupuis (approved):

PR #678, opened 2021-07-15
feat/link-preview-styles

hchevallier:
# General summary
This PR adds ellipsis on title and link.

# Screenshots

# Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   TextField: Add space after asterisk if component has `isRequired` props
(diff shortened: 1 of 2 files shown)


matmkian (changes requested):
I'm not sure we should do this... Force ellipsis on plink previews?

hchevallier:
> I'm not sure we should do this... Force ellipsis on plink previews? Or maybe not this way...

Actually, this is already on one line, and it's truncated for now, without ellipsis. And Articles Preview mockups are showing ellipsis on the link (not on the title, but since it was also forced on one line, I assume it should be ellipsed instead of truncated)

matmkian:
Ok let's talk about that @hchevallier, we should do that in a different way...

matmkian:
The element itself should have specific behavior regarding the context... Less code and more understandable.

PR #685, opened 2021-08-12
feat(switch): add role switch to html input for screen reader

aureldent:
# General summary
This PR adds a role to the switch input element (checkbox) to enable screen readers to know that's a switch.
This will make is so the screen reader will tell `switch` instead of `checkbox`

Idea comes from this video: https://youtu.be/_KqccADghcA?t=176
Or this article: https://web.dev/building-a-switch-component/

# Screenshots

# Check list

-   [] (if has style) Add `need: review-integration` label


pidupuis (approved):

gcornut (changes requested):
@aureldent Thank you for your PR, the code looks great but we'll need the following:
1. The PR should be based on `release/v1`  instead of `master` (that's temporary until we release v2)
2. Add an entry in the `CHANGELOG.md` (in "Unreleased" > "Added")

PR #686, opened 2021-08-23
feat(switch): add role switch for screen readers behavior

aureldent:
# General summary
This PR adds a role to the switch input element (checkbox) to enable screen readers to know that's a switch.
This will make is so the screen reader will tell `switch` instead of `checkbox`

Idea comes from this video: https://youtu.be/_KqccADghcA?t=176
Or this article: https://web.dev/building-a-switch-component/

# Screenshots

# Check list

-   [] (if has style) Add `need: review-integration` label

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


gcornut (approved):

gcornut:
This will be available in the next v1 patch release. No release date scheduled for now, we'll try to add other fixes or minor improvement before a new path release

PR #711, opened 2021-11-04
feat(button): allow button to take available width

dbenfouzari:
# General summary

Add a `fullWidth` prop on `Button` to make it as big as available

# Screenshots

image

<details>
<summary>Bonus: added stories for all button/icon button variations</summary>

html_id=lumx-components-button-button--button-variations

html_id=lumx-components-button-button--icon-button-variations

</details>

# Check list

-   [X] (if has style) Add `need: review-integration` label
-   [X] (if has textual documentation) Add `need: review-doc` label
-   [X] (if has code) Add `need: review-frontend` label
-   [X] (if has react) Check through the [react dev check list]
-   [X] (if fix or feature) Add `need: test` label

Check out the [contribution guide] for general guidelines for submissions to the design system.

[contribution guide]: https://github.com/lumapps/design-system/blob/master/CONTRIBUTING.md
[react dev check list]: https://github.com/lumapps/design-system/blob/master/CONTRIBUTING.md#react-developpment-check-list

PR #712, opened 2021-11-04
wip(button): fix button hasBackground markup

gcornut:
# General summary

- [x] Remove button wrapper to use a ::after pseudo element instead
- [ ] Fix the hover background color


gcornut:
Won't do, requires a full rework of the mixins. For now the button wrapper width bug will be fixed in https://github.com/lumapps/design-system/pull/711

PR #722, opened 2021-11-16
chore(storybook): update to v6 and add a material theme switcher

gcornut:
- Isolate storybook dependencies (move it out of the yarn workspace to avoid conflict with the demo site dependencies)
- Update to storybook v6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


FlorentDinet:
👏 

Same minor feedback as for the site
I'd use vanilla switches instead of label-customized ones as they appear not aligned vertically
Capture d’écran 2021-11-17 à 15 16 27
:

gcornut:
@FlorentDinet I'm moving the material them switcher for both storybook and the demo site in a single PR: https://github.com/lumapps/design-system/pull/723 
It should simplify things :)

PR #724, opened 2021-11-17
feat(list-item): add isDisabled prop

gcornut:
# General summary

Add disabled state on the list item

### DIFF
--- a/packages/lumx-react/src/components/list/ListItem.tsx
+++ b/packages/lumx-react/src/components/list/ListItem.tsx
@@ -23,6 +23,8 @@ export interface ListItemProps extends GenericProps {
     isHighlighted?: boolean;
     /** Whether the component is selected or not. */
     isSelected?: boolean;
+    /** Whether link/button is disabled or not. */
+    isDisabled?: boolean;
     /** Reference to the <li> element. */
     listItemRef?: Ref<HTMLLIElement>;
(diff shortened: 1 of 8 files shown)


FlorentDinet:
👏 
I've noted to myself to improve the doc to mention selected and disabled states

juanigalan91 (approved):
LGTM

packages/lumx-react/src/components/list/ListItem.tsx
```diff
@@ -115,15 +119,18 @@ export const ListItem: Comp<ListItemProps, HTMLLIElement> = forwardRef((props, r
                         linkAs,
                         tabIndex: 0,
                         role: onItemSelected ? 'button' : undefined,
+                        'aria-disabled': isDisabled,
                         ...linkProps,
+                        href: isDisabled ? undefined : linkProps.href,
```
Mcourt, line 124: Nice. 
I wasn't sure about something and found those two articles: 
https://www.scottohara.me/blog/r2021/05/28/disabled-links.html
https://a11y-guidelines.orange.com/en/articles/disable-elements/

I initially thought you couldn't set a "aria-disabled" on a link and having an empty href was enough, but it looks like it's fine **as long** as you also add `role="link"` (but only when it's a link and it's disabled).

PR #775, opened 2022-02-07
chore:(input): pass props to checkbox and switch

juanigalan91:
# General summary
In cases where we need to manage the label of checkboxes and switches from outside the actual component, we need to pass props into the input to tell it the id of element for that label.

This PR adds an `inputProps` prop for `Checkbox` and `Switch` so that we can pass in props such as `aria-labelledby`, which will allow us to manage the label of a given checkbox or switch from outside the DS component.

# Check list


gcornut (changes requested):
Can you add an entry in the CHANGELOG (under: `## Unreleased > ### Added`) ?

gcornut (approved):

PR #786, opened 2022-03-02
chore(slideshow): improve aria labels and focusable elements

juanigalan91:
# General summary
First batch of fixes for the slideshow when it comes to accessibility:
- Make controls focusable
- Add several aria roles and description
- Stop autoplay on mouse enter and on focus

# Screenshots

# Check list

### DIFF
--- a/packages/lumx-react/src/components/slideshow/Slideshow.tsx
+++ b/packages/lumx-react/src/components/slideshow/Slideshow.tsx
@@ -1,13 +1,14 @@
-import React, { CSSProperties, forwardRef, useCallback, useEffect, useState } from 'react';
+import React, { CSSProperties, forwardRef, useCallback, useEffect, useMemo, useState } from 'react';
 
 import classNames from 'classnames';
 
 import { SlideshowControls, SlideshowControlsProps, Theme } from '@lumx/react';
-
 import { AUTOPLAY_DEFAULT_INTERVAL, FULL_WIDTH_PERCENT } from '@lumx/react/components/slideshow/constants';
 import { useInterval } from '@lumx/react/hooks/useInterval';
 import { Comp, GenericProps, getRootClassName, handleBasicClasses } from '@lumx/react/utils';
 import { mergeRefs } from '@lumx/react/utils/mergeRefs';
+
+import uniqueId from 'lodash/uniqueId';
 
 /**
  * Defines the props of the component.
@@ -39,6 +40,10 @@ export interface SlideshowProps extends GenericProps {
     theme?: Theme;
     /** Callback when slide changes */
     onChange?(index: number): void;
+    /** slideshow HTML id attribute */
+    id?: string;
+    /** slides wrapper HTML id attribute */
+    slidesId?: string;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/lumx-react/src/components/slideshow/Slideshow.tsx
```diff
@@ -182,18 +182,29 @@ export const Slideshow: Comp<SlideshowProps, HTMLDivElement> = forwardRef((props
...
+            aria-live={isAutoPlaying ? 'off' : 'polite'}
         >
-            <div className={`${CLASSNAME}__slides`}>
+            <div
+                id={slideshowSlidesId}
+                className={`${CLASSNAME}__slides`}
+                onMouseEnter={() => setIsAutoPlaying(false)}
+                onMouseLeave={() => setIsAutoPlaying(Boolean(autoPlay))}
+                onBlur={() => setIsAutoPlaying(Boolean(autoPlay))}
+                onFocus={() => setIsAutoPlaying(false)}
```
Mcourt, line 205: This is where it may get a bit complicated (and can be tackled on a later PR), but this behavior should be applied to:
* The slide itself, as you did here
* Any interactive element within the slide, which may be a bit tricky...
* The controls (except the future pause button, if we follow the WAI guidelines).

Maybe we could add some sort of 'focusWithin' behavior ? :thinking:

packages/lumx-react/src/components/slideshow/Slideshow.tsx
```diff
@@ -182,18 +182,29 @@ export const Slideshow: Comp<SlideshowProps, HTMLDivElement> = forwardRef((props
...
+                tabIndex={0}
```
Mcourt, line 206: I don't think the whole slide should be focusable by default.
It should only be focusable if it is interactive (button / link).

Maybe we could "simply" let the child decide of this behavior ? 
Don't set any tabIndex at all and whatever is set inside the slideshow will decide if it should be focusable or not.

Mcourt (approved):
LGTM. 

Tested with a screen reader and it's a **huge** improvements.
As discussed there are a few areas that could be further improved here and there, but this is already great!  

Kudos!

gcornut (approved):

packages/lumx-react/src/components/slideshow/Slideshow.tsx
```diff
@@ -182,20 +182,69 @@ export const Slideshow: Comp<SlideshowProps, HTMLDivElement> = forwardRef((props
         onChange(currentIndex);
     }, [currentIndex, onChange]);
 
+    const slideshowSlidesId = 'slideshow-slides';
+    const slideshowId = 'slideshow';
```
gcornut, line 186: This will actually be a problem when there is multiple slideshow on the same page. 
I think we should make these customizable via props so that we can use the slideshow controls separated from the slideshow but still connected via `aria-controls`
I would propose:
```js
const { id, slidesId } = props;

const slideshowId = useMemo(() => id || uniqId('slideshow'), [id]) 
const slideshowSlidesId = useMemo(() => slidesId || uniqId('slideshow-slides'), [slidesId]) 
```

packages/lumx-react/src/components/slideshow/Slideshow.tsx
```diff
@@ -182,20 +182,69 @@ export const Slideshow: Comp<SlideshowProps, HTMLDivElement> = forwardRef((props
...
+
+    useEffect(() => {
+        const slideshow = document.getElementById('slideshow');
```
gcornut, line 189: We can use the `element` state directly, no need to get the element.

```suggestion
        if (!element) return
        const slideshow = element;
```

gcornut (changes requested):

gcornut (approved):

PR #792, opened 2022-03-09
chore(slideshow): create hook for slideshow controls

juanigalan91:
# General summary
Refactor the Slideshow controls management so that we can reuse the logic if the components Slideshow and SlideshowControls are used separately.

# Screenshots


gcornut (approved):

gcornut:
Test OK in storybook

PR #793, opened 2022-03-09
WIP export hook in @lumx/react build

gcornut:
# General summary


gcornut:
Reworked and simplified here https://github.com/lumapps/design-system/pull/796

PR #799, opened 2022-03-16
feat(icon): add dark theme compatibility

FlorentDinet:
# General summary

Icon component do not currently support dark theme, this work is my attempt to make icon with "hasShape" parameter supports dark theme.

# Screenshots

What I'm trying to achieve:
Capture d’écran 2022-03-16 à 10 35 52

I'm available to discuss the case.

PR #804, opened 2022-03-17
chore(icon): add variants in stories

gcornut:


gcornut:
Merged in https://github.com/lumapps/design-system/pull/799

PR #805, opened 2022-03-18
Fix/dropdown focus improvements

Mcourt:
# General summary

* Removed default `tabIndex={0}` on lists and added a `tabIndex` prop default at `-1`
* Made the `Popover` element focus back on it's anchor when closed while focus is within it. Controllable via `focusAnchorOnClose`, default to `true`.
* Added `focusAnchorOnClose` on `Dropdown` element to forward it to inner `Popover`. Default to `true`.

# Screenshots
focusOnAnchor


packages/lumx-react/src/components/list/List.tsx
```diff
@@ -21,6 +21,8 @@ export interface ListProps extends GenericProps {
     isClickable?: boolean;
     /** Item padding size. */
     itemPadding?: Extract<Size, 'big' | 'huge'>;
+    /** Tab index of the list. Default to -1 */
+    tabIndex?: -1 | 0;
```
gcornut, line 25: It's technically a breaking change to reduce the tabIndex to `0 | -1`
```suggestion
    tabIndex?: number;
```

CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Popover: Added `focusAnchorOnClose` prop to reset the focus back to the anchor element when popover closes and focus is set within. Default to `true`;
+- Dropdown: Added `focusAnchorOnClose` prop to reset the focus back to the anchor element when popover closes and focus is set within. Default to `true`;
+- List: Added `tabIndex` prop to control the list tabIndex. Default to `-1` to avoid breaking changes.
```
gcornut, line 14: Why is it avoiding breaking change to default to `-1` instead of `undefined` ?

gcornut (changes requested):

CHANGELOG.md
Mcourt: Because I believe some components, like the Dropdown I think, sets the focus on the List when it appears.

If the tabIndex is undefined, this simply wont work. You need to add `tabIndex={-1}` in order to programmatically set a focus on a element.

packages/lumx-react/src/components/list/List.tsx
Mcourt: True, will change

packages/lumx-react/src/components/popover/Popover.tsx
```diff
@@ -222,6 +226,38 @@ export const Popover: Comp<PopoverProps, HTMLDivElement> = forwardRef((props, re
     // eslint-disable-next-line react-hooks/rules-of-hooks
     const clickAwayRef = useRef<HTMLDivElement>(null);
 
+    /**
+     * Track whether the focus is currently set in the
+     * popover.
+     * */
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const [isFocusedWithin, setIsFocusedWithin] = useState(false);
```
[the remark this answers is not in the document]
Mcourt: Sure.

PR #807, opened 2022-03-21
chore(tooltip): remove unecessary set state on unmount

gcornut:
# General summary

Removed unecessary closing of tooltip (react set state) on unmount that can cause warnings in the React dev tools


gcornut:
Test OK - no regression on tooltip close

juanigalan91 (approved):

Mcourt (approved):

PR #812, opened 2022-04-08
chore(comment-block): change date position and add tooltip for date

Fafnor:
# General summary

Move the block comment date to below comment text and add a tooltip to display the new full date prop because the date will used the new timestamp format displaying the minimal information needed according to the current date

# Screenshots
image


gcornut (approved):

gcornut:
Test OK, date displayed at the end and full date on hover :heavy_check_mark:

PR #822, opened 2022-04-26
chore(a11y): fix focus trap not ignoring disabled elements

gcornut:
# General summary

2 focus trap issues fixed in this PR:
- Having disabled elements inside dialogs and lightbox could break the focus trap.
- Also coming in focus from outside the browser window could escape the focus trap.

# Test

Test on storybook: https://5fbfb1d508c0520021560f10-xzwarxjqom.chromatic.com/iframe.html?id=lumx-components-dialog-dialog--dialog-with-focusable-elements&args=&viewMode=story
The dialog should trap the focus
- Using Tab or using shift+Tab should circle through all focusable elemnts in the dialog without leaving the dialog
- Starting the focus in the browser search bar and focusing in the window should focus inside the dialog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


Mcourt (approved):
LGTM

PR #823, opened 2022-04-26
feat(size): add medium

armasson:
# General summary

Add the 'medium' option inside the available Size map and GapSize type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Dialog & Lightbox: fixed focus trap on disabled elements and protect against focus breaking out.
 
+### Added
+
+-   Size map: added `medium`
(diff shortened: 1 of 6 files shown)


gcornut (changes requested):
Can you add a CHANGELOG entry ? 
Also we could test the FlexBox gap size in storybook, can you add a story or add a knob in the `Flex` story ? We'll deploy a storybook and test on it

armasson:
> Can you add a CHANGELOG entry ? Also we could test the FlexBox gap size in storybook, can you add a story or add a knob in the `Flex` story ? We'll deploy a storybook and test on it

I've updated the stories and test with a Knobs and a little story for the GapSize. And by doing that I realised that I had to update the scss too 😅

gcornut (approved):
Looks good, I'll take the test on storybook

gcornut:
Test OK, felxbox gap size medium working as expected

PR #836, opened 2022-05-31
fix(user-block): improve a11y keyboard nav

gcornut:
- fix(user-block): improve a11y keyboard navigation
- chore(demo-site): document user-block accessibility

Remove keyboard focus on the avatar to avoid redundancy with the name link/button


Mcourt (approved):
LGTM

PR #839, opened 2022-06-01
fix(tooltip): fix tooltip render in SSR

gcornut:
# General summary

Fix tooltip render in SSR that should render its children instead of null.

## Test:
- Build the demo site (`cd packages/site-demo && yarn clean && yarn build`)
- Run static HTTP server (`npx http-server public`)
- Load the website in responsive mode and check the menu button is visible on first render (rendered in static HTML) and does not produce a layout shift


juanigalan91 (approved):

gcornut:
Test OK
(fixed the header wrapping on small screens)

PR #840, opened 2022-06-03
fix(a11y): fix focus trap ignoring element with disabled=false

gcornut:
# General summary

Elements with `disabled=false` or `aria-disabled=false` at the end or start of a dialog or lightbox were skipped in the focus trap.

This PR fixes that and adds tests on the code detecting the first and last focusable elements in dialogs and lightbox.

## Test:
- [Go to the dialog focus trap story
](https://5fbfb1d508c0520021560f10-efpjqlzlmv.chromatic.com/?path=/story/lumx-components-dialog-dialog--dialog-focus-trap)
- Check that the focus is contained inside the dialog
- Every focusable item is reachable
- Every non focusable item are skipped

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


gcornut:
Test OK
Button with disabled=false at the end of a dialog is focusable

PR #844, opened 2022-06-08
docs(button): document accessibility concerns

hchevallier:
# General summary

Adds some documentation about accessibility for the `<Button>` component

# Screenshots

### DIFF
--- a/packages/site-demo/content/product/components/button/index.mdx
+++ b/packages/site-demo/content/product/components/button/index.mdx
@@ -40,6 +40,27 @@ Use full width when you need to get the button full width.
 
 <DemoBlock vAlign="center" demo="full-width" withThemeSwitcher />
 
+### Accessibility concerns
+
+#### Button
+
+Button will use the text within the button as the accessible name, if there is no text in it (replaced by an image, an icon, etc ...) then you **should pass an `aria-label`** describing what the button does.
+
+Buttons may have 3 specific accessibility use cases:
+
+- **Button as a link** if you set the `linkAs` or the `href` prop, then your button will use the `<a>` tag.
+In this case, the label of the button (or `aria-label`) will be enough to make your link accessible.
+- **Toggle button**, when your button as an active/inactive state (`isSelected` prop is responsible of this state). 
+If the label of the button **changes depending on the state, then it doesn't need anything more.**
+But if the label of the button **doesn't change you should add a `aria-pressed=true/false`** to specify to the user in which state the button is.
+ - **Menu button** if your button opens a dropdown. 
+ You have to set, on the button, `aria-haspopup` to `true`, and `aria-expanded` to `true` when the dropdown is open.
+ Also, when the dropdown is open, you have to add `aria-controls` on the button, referencing the id of the dropdown. If your dropdown is not present in the DOM when it is closed, then you should not provide the `aria-controls` prop.
+
+#### IconButton
+
+All concerns about the `Button` component are true for `IconButton`, but as there is no text in the button, you have to provide a `label` describing what the button does.
+
 ### Button properties
 
 <PropTable component="Button" />
--- a/packages/lumx-react/src/components/button/Button.stories.tsx
+++ b/packages/lumx-react/src/components/button/Button.stories.tsx
@@ -13,6 +13,7 @@ const DEFAULT_PROPS = Button.defaultProps as any;
 export const SimpleButton = ({ theme }: any) => {
     return (
         <Button
+            aria-pressed={boolean('isSelected', Boolean(DEFAULT_PROPS.isSelected))}
             emphasis={emphasis('Emphasis', DEFAULT_PROPS.emphasis)}
             theme={theme}
             rightIcon={select('Right icon', { none: undefined, mdiSend }, undefined)}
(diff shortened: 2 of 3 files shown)


packages/site-demo/content/product/components/button/index.mdx
```diff
@@ -40,6 +40,26 @@ Use full width when you need to get the button full width.
...
+You **should define the `alt` attribute**, with a string indicating where the link will take the user.
+- **Toggle button**, when your button as an active/inactive state (`isSelected` prop is responsible of this state). 
+If the label of the button **changes depending on the state, then it doesn't need anything more.**
+But if the label of the button **doesn't change you should add a `aria-pressed=true/false`** to specify to the user in which state the button is.
+ - **Menu button** if your button opens a dropdown. 
+ You have to set, on the button, `aria-haspopup` to `true`, and `aria-expanded` to `true` when the dropdown is open.
+
+#### IconButton
+
+All concerns about the `Button` component are true for `IconButton`, but as there is no text in the button, you have to provide an `aria-label` describing what the button does.
```
gcornut, line 61:
```suggestion
All concerns about the `Button` component are true for `IconButton`, but as there is no text in the button, you have to provide a `label` describing what the button does.
```

gcornut (changes requested):
Can we document the props `aria-pressed`, `aria-haspopup` and `aria-expanded` in the base button props JSDoc?

packages/site-demo/content/product/components/button/index.mdx
```diff
@@ -40,6 +40,26 @@ Use full width when you need to get the button full width.
...
+
+Buttons may have 3 specific accessibility use cases:
+
+- **Button as a link** if you set the `linkAs` or the `href` prop, then your button will use the `<a>` tag.
+You **should define the `alt` attribute**, with a string indicating where the link will take the user.
+- **Toggle button**, when your button as an active/inactive state (`isSelected` prop is responsible of this state). 
+If the label of the button **changes depending on the state, then it doesn't need anything more.**
+But if the label of the button **doesn't change you should add a `aria-pressed=true/false`** to specify to the user in which state the button is.
+ - **Menu button** if your button opens a dropdown. 
+ You have to set, on the button, `aria-haspopup` to `true`, and `aria-expanded` to `true` when the dropdown is open.
```
gcornut, line 57: "Menu" is a very specific ARIA pattern, should we be more general and describe the popup use case (menu, combobox, dialog, etc.) and separately describe the expanded use case (menu, disclosure, popup, etc.) ?
Also it looks like `aria-expanded` should always be paired with `aria-controls` or `aria-owns`

gcornut (changes requested):

packages/site-demo/content/product/components/button/index.mdx
hchevallier: I took the [`Action Menu Button`](https://w3c.github.io/aria-practices/examples/menu-button/menu-button-actions.html) as a reference, this use case is listed in the [Button spec](https://w3c.github.io/aria-practices/#button)
I'm not sure what you mean, for dialog pattern I think it should be specified in the Dialog component. Maybe we can add here some links to the different DS components docs, to list all specific patterns 

`aria-controls` can't refer to an element that is not in the DOM, not sure this is relevant in our case, but I may be wrong

hchevallier: Ok just read that: 
> The aria-controls only needs to be set when the popup is visible, but it is valid and easier to program to reference an element that is not visible.
I'll add this

hchevallier: > Also, when the dropdown is open, you have to add `aria-controls` on the button, referencing the id of the dropdown. 
 If your dropdown is not present in the DOM when it is closed, then you should not provide the `aria-controls` prop.

 Does it looks ok for you?

gcornut: Yep, it should be sufficient

packages/site-demo/content/product/components/button/index.mdx
```diff
@@ -40,6 +40,27 @@ Use full width when you need to get the button full width.
...
+Buttons may have 3 specific accessibility use cases:
+
+- **Button as a link** if you set the `linkAs` or the `href` prop, then your button will use the `<a>` tag.
+In this case, the label of the button (or `aria-label`) will be enough to make your link accessible.
+- **Toggle button**, when your button as an active/inactive state (`isSelected` prop is responsible of this state). 
+If the label of the button **changes depending on the state, then it doesn't need anything more.**
+But if the label of the button **doesn't change you should add a `aria-pressed=true/false`** to specify to the user in which state the button is.
+ - **Menu button** if your button opens a dropdown. 
+ You have to set, on the button, `aria-haspopup` to `true`, and `aria-expanded` to `true` when the dropdown is open.
+ Also, when the dropdown is open, you have to add `aria-controls` on the button, referencing the id of the dropdown. If your dropdown is not present in the DOM when it is closed, then you should not provide the `aria-controls` prop.
```
gcornut, line 58: We'll look into making such a pattern trivial with a new Menu/MenuTrigger component

gcornut (approved):

packages/site-demo/content/product/components/button/index.mdx
hchevallier: I agree

PR #849, opened 2022-06-27
chore(focus-trap): fix & improve the focus trap

gcornut:
- fix(focus-trap): fix reset focus trap
- fix(focus-trap): fix nested focus traps
- fix(focus-trap): fix close nested components on escape key pressed
- chore(popover): focus first focusable element inside anchor on close
- fix(date-picker-field): fixed focus element on open

# Test

Scenario:
1. Go to the story ["Dialog Focus Trap"](https://5fbfb1d508c0520021560f10-enyqcdxplj.chromatic.com/?path=/story/lumx-components-dialog-dialog--dialog-focus-trap)
1. Navigate with `Tab` to the "Open date picker" button
1. Open by pressing `Enter`
1. Navigate with `Tab` to the date picker field
1. Open by pressing `Enter`
1. Navigate dates with `Tab` (this navigation could be improved upon)
1. Close the date picker with `Escape`
1. Close the date picker dialog with `Escape`
1. Close the main dialog with `Escape`

Additional checks 
- Navigating with `Tab` or `shift+Tab` should be confined to the dialog/date-picker
- Closing a dialog/date-picker should focus the activating element (button or date-picker field)
- Opening a dialog/date-picker should focus an element by default

### DIFF
--- a/packages/lumx-react/src/hooks/useCallbackOnEscape.ts
+++ b/packages/lumx-react/src/hooks/useCallbackOnEscape.ts
@@ -1,25 +1,54 @@
 import { DOCUMENT } from '@lumx/react/constants';
 import { Callback, onEscapePressed } from '@lumx/react/utils';
 import { useEffect } from 'react';
+import last from 'lodash/last';
+import pull from 'lodash/pull';
+
+type Listener = { enable(): void; disable(): void };
 
 /**
- * Triggers a callback when the escape key is pressed.
+ * List of all listeners.
+ */
+const LISTENERS: Listener[] = [];
+
+/**
+ * Register a global listener on 'Escape' key pressed.
+ *
+ * If multiple listener are registered, only the last one is maintained. When a listener is unregistered, the previous
+ * one gets activated again.
  *
  * @param callback      Callback
  * @param closeOnEscape Disables the hook when false
- * @param rootElement   Element on which to listen the escape key
  */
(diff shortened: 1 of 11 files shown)


mathieubrisson:
Hello @matmkian 

Could you have a look to this PR please ?
Thank you

packages/lumx-react/src/hooks/useCallbackOnEscape.ts
```diff
@@ -1,25 +1,54 @@
...
+        // TEARDOWN:
+        return () => {
+            // Disable current listener.
+            listener.disable();
+            // Remove current listener.
+            pull(LISTENERS, listener);
+            // Enable previous listener.
+            last(LISTENERS)?.enable();
+        };
+    }, [callback, closeOnEscape]);
```
Mcourt, line 53: I might be wrong, I don't have the full context and constraints in mind, but this code looks a lot like what you added in `useFocusTrap`.

Is there anyway we could reuse it somehow? Or maybe we voluntarily don't want to?

gcornut: Good catch, I intended to mutualized but didn't find own to name this pattern I forgot to do it

gcornut:
Rebased & factorized code

Mcourt:
> Rebased & factorized code

Beautiful :')

PR #850, opened 2022-06-28
chore(link-preview): minor a11y and markup improvements

gcornut:
# General summary

LinkPreview
- Added `linkAs` prop to customize link component
- Improve markup (use `<article` at the root and `<h2` for the title)
- Keep only one link in focus order (others have `tabindex=-1`)

# Test

Go to storybook: https://5fbfb1d508c0520021560f10-kgvsyblzex.chromatic.com/?path=/story/lumx-components-link-preview-link-preview--regular-link-preview

Check on all link preview stories:
- Inspect the DOM
  - Link preview are wrapped in `<article`
  - Title should be wrapped in `<h2`
  - Thumbnail, title and link should all be wrapped into a link with `target=_blank`
- Only one link should be focusable with Tab (title or link if title is not present)


packages/lumx-react/src/components/link-preview/LinkPreview.tsx
```diff
@@ -95,35 +111,40 @@ export const LinkPreview: Comp<LinkPreviewProps, HTMLDivElement> = forwardRef((p
 
                 <div className={`${CLASSNAME}__container`}>
                     {title && (
-                        <div className={`${CLASSNAME}__title`}>
+                        <h2 className={`${CLASSNAME}__title`}>
```
Mcourt, line 114: I'm not too sure about forcing this to be a h2.
As this is a highly reusable component, it would break the hiearchy to have it beneath an h3 for example.

Could we have a prop to choose the title level?

packages/lumx-react/src/components/link-preview/LinkPreview.tsx
```diff
@@ -62,13 +67,24 @@ const DEFAULT_PROPS: Partial<LinkPreviewProps> = {
...
+        linkProps,
+        size,
+        theme,
+        thumbnailProps,
+        title,
+        titleHeading,
+        ...forwardedProps
+    } = props;
+    // Use title heading as title wrapper (see DEFAULT_PROPS for the default value).
+    const TitleHeading = titleHeading as HeadingElement;
```
Mcourt, line 84: Huh, was the `as` needed?
(Not an issue, just genuinely curious)

Mcourt (approved):
LGTM  :) Great !

packages/lumx-react/src/components/link-preview/LinkPreview.tsx
gcornut: Yeah it's needed otherwise the type is `HeadingElement | undefined` since he does not now that React will use the `defaultProps` if the prop is undefined and thus this prop cannot be `undefined`

Mcourt: Ah I see, very well then :)

PR #858, opened 2022-07-20
chore(doc): factorize components theme prop documentation

gcornut:
Factorize theme prop documentation across components

All component having `theme` props extend the `HasTheme` interface on which the `theme` prop is documented with JSDoc.

On the documentation site, the `theme` prop will show in all component prop tables

### DIFF
--- a/packages/lumx-react/src/components/autocomplete/Autocomplete.tsx
+++ b/packages/lumx-react/src/components/autocomplete/Autocomplete.tsx
@@ -2,9 +2,9 @@ import React, { forwardRef, ReactNode, SyntheticEvent, useRef } from 'react';
 
 import classNames from 'classnames';
 
-import { Dropdown, IconButtonProps, Offset, Placement, TextField, TextFieldProps, Theme } from '@lumx/react';
+import { Dropdown, IconButtonProps, Offset, Placement, TextField, TextFieldProps } from '@lumx/react';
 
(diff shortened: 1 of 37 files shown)


carlo-lumapps (approved):
Thanks!

packages/lumx-react/src/components/button/IconButton.test.tsx
```diff
@@ -79,7 +79,7 @@ describe(`<${IconButton.displayName}>`, () => {
             const { wrapper, buttonRoot } = setup(props);
             expect(wrapper).toMatchSnapshot();
 
-            expect(buttonRoot).toHaveClassName(props.className);
+            expect(buttonRoot).toHaveClassName(props.className as string);
```
radiuso, line 82: Why do you need to cast the className here?

radiuso (approved):

packages/lumx-react/src/components/button/IconButton.test.tsx
gcornut: Because `HasClassName` has `className?: string`. The `className` is always optional (previously it had the implicit type `any` and thus the cast was not necessary)

gcornut:
No impact on functional code (only documentation & types)
Non regression test OK (demo site): theme prop documentation appears in properties table :heavy_check_mark:

PR #859, opened 2022-07-22
fix(click-away): fix nested click away context

gcornut:
# General summary

Fix nested click away handling.

Nested click away contexts should add up so that children click away context are considered as part of the parent click away context.

Dialogs are special in that clicking outside a dialog should still be considered as clicking inside the parent dialogs.

## Test scenario 1

A custom scenario with no real application (not using exported DS components):
1. Go to the story: https://5fbfb1d508c0520021560f10-futsykytnj.chromatic.com/?path=/story/lumx-components-clickawayprovider--nested-click-away
2. Open all nested click away contexts
3. Clicking inside one should close the children level and keep the parent level open
4. Repeat test at all levels

## Test scenario 2: Dialogs

1. Go to story: https://5fbfb1d508c0520021560f10-futsykytnj.chromatic.com/?path=/story/lumx-components-dialog-dialog--dialog-focus-trap
2. A dialog should be open 
3. Inside of this dialog, click on the "Open date picker" button
4. Another dialog should open on top of the previous one  
5. Inside of this dialog, click on a date picker field, it should open in dropdown under
6. Now click outside the dialog. The date picker dropdown and the date picker dialog should both close but the parent dialog should not close
7. Now click outside the remaining dialog. It should close normally


charles-lumapps (approved):
Thank you!

carlo-lumapps (approved):
It looks good to me. Thanks a lot 🙏

PR #862, opened 2022-07-26
feat/generic block improvements

gcornut:
- feat(generic-block): improve default prop values
- feat(generic-block): remove sections if empty
- feat(generic-block): prevent content overflow by default

### DIFF
--- a/packages/lumx-react/src/components/generic-block/GenericBlock.stories.tsx
+++ b/packages/lumx-react/src/components/generic-block/GenericBlock.stories.tsx
@@ -49,6 +49,22 @@ export const Vertical = ({ theme }: HasTheme) => (
     </GenericBlock>
 );
 
+export const VerticalOverflow = ({ theme }: HasTheme) => (
+    <div style={{ width: 300 }}>
+        <GenericBlock orientation={Orientation.vertical}>
+            <GenericBlock.Figure style={redBorderStyle}>
+                <Icon icon={mdiPencil} size={Size.m} />
+            </GenericBlock.Figure>
+            <GenericBlock.Content style={redBorderStyle}>
+                _________________________________________________________________
+            </GenericBlock.Content>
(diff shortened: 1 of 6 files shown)


packages/lumx-react/src/components/generic-block/GenericBlock.stories.tsx
```diff
@@ -49,6 +49,22 @@ export const Vertical = ({ theme }: HasTheme) => (
...
+export const VerticalOverflow = ({ theme }: HasTheme) => (
+    <div style={{ width: 300 }}>
+        <GenericBlock orientation={Orientation.vertical}>
+            <GenericBlock.Figure style={redBorderStyle}>
+                <Icon icon={mdiPencil} size={Size.m} />
+            </GenericBlock.Figure>
+            <GenericBlock.Content style={redBorderStyle}>
+                _________________________________________________________________
+            </GenericBlock.Content>
+            <GenericBlock.Actions fillSpace style={redBorderStyle}>
```
Mcourt, line 61: Do we want the actions to `fillSpace` by default? It looks like they always are in the stories.

Mcourt (approved):
Small comment but otherwise LGTM

packages/lumx-react/src/components/generic-block/GenericBlock.stories.tsx
gcornut: No actually here `fillSpace` does not have any effect in vertical mode without height constraint ([see storybook](https://5fbfb1d508c0520021560f10-tcmubuhvry.chromatic.com/?path=/story/lumx-components-generic-block-genericblock--vertical-overflow)) I'll remove it.
It is activated by default on the content section though

Mcourt: Ah I see. Anyway, it was nitpicking :) Looks good to me.

gcornut:
Test OK (100% coverage on unit test)

PR #864, opened 2022-07-28
feat(text): create text component

gcornut:
# General summary

Create `Text` component.

### DIFF
--- a/configs/postcss.config.js
+++ b/configs/postcss.config.js
@@ -1,7 +1,6 @@
 /* eslint-disable global-require, import/no-commonjs, import/unambiguous */
 module.exports = {
     plugins: [
-        require('postcss-flexbox-unboxer')(),
         require('autoprefixer')({
             grid: true,
         }),
--- a/packages/lumx-react/src/components/text/Text.tsx
+++ b/packages/lumx-react/src/components/text/Text.tsx
@@ -1,7 +1,7 @@
 import React, { forwardRef } from 'react';
 
 import { Color, ColorVariant, Typography } from '@lumx/react';
-import { Comp, GenericProps, getRootClassName, HeadingElement } from '@lumx/react/utils';
+import { Comp, GenericProps, getRootClassName, handleBasicClasses, HeadingElement } from '@lumx/react/utils';
 import classNames from 'classnames';
 
 type TextComponents = 'span' | 'p' | HeadingElement;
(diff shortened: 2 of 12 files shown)


packages/lumx-react/src/components/text/Text.tsx
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Component default class name and class prefix.
+ */
+const CLASSNAME = getRootClassName(COMPONENT_NAME);
+
+/**
+ * Component default props.
+ */
+const DEFAULT_PROPS = {
+    as: 'span',
```
Mcourt, line 51: This actually was my biggest gripe about the component.
IMO the default element should be a `p` and not a `span`, for accessible reasons.

Setting it as `span` by default means we end up with components looking like this 

```jsx
<article>
  <h2>My Title</h2>
  <Text>The description of the article</Text>
</article>
```

Which will render the following html 

```html
<article>
  <h2>My Title</h2>
  <span>The description of the article</span>
</article>
```

which is neither valid nor accessible html.

Unless there are some usecases I can't think of right now, a `span` component should only be used within an element that already has semantics (headings / paragraphs), for styling purposes (or to add some other attribute).

Which means we need to add `as="p"` more times than we would need to ad `as="span"`.

configs/postcss.config.js
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable global-require, import/no-commonjs, import/unambiguous */
 module.exports = {
     plugins: [
-        require('postcss-flexbox-unboxer')(),
```
Mcourt, line 4: Was that added in this PR on purpose?

packages/lumx-react/src/components/text/Text.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { forwardRef } from 'react';
+
+import { Color, ColorVariant, Typography } from '@lumx/react';
+import { Comp, GenericProps, getRootClassName, handleBasicClasses, HeadingElement } from '@lumx/react/utils';
+import classNames from 'classnames';
+
+type TextComponents = 'span' | 'p' | HeadingElement;
```
Mcourt, line 7: Should we allow more components? Like `strong` / `em` / etc...

configs/postcss.config.js
gcornut: Yes it's required otherwise the `-webkit-box` is removed and we can't do multiline truncate. Also it has be removed from the product too a while ago

Mcourt (approved):

PR #865, opened 2022-07-28
chore(icons): build module for better tree shaking

gcornut:
# General summary

Add a build step to the `@lumx/icons` to flatten all icons from mdi into one ESM module so tree shaking can be more effective

### DIFF
--- a/packages/lumx-icons/rollup.config.js
+++ b/packages/lumx-icons/rollup.config.js
@@ -1,0 +1,42 @@
+import path from 'path';
+import glob from 'glob';
+import resolve from '@rollup/plugin-node-resolve';
+import babel from 'rollup-plugin-babel';
+import copy from 'rollup-plugin-copy';
+import cleaner from 'rollup-plugin-cleaner';
+
+import pkg from './package.json';
+
+const CONFIGS = require('../../configs');
+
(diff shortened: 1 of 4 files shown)


packages/lumx-icons/rollup.config.js
```diff
@@ -0,0 +1,47 @@
...
+            extensions,
+            exclude: /node_modules/,
+            plugins: CONFIGS.babel.plugins,
+            presets: [
+                ['@babel/preset-env', { targets: 'defaults' }],
+            ],
+        }),
+        /** Copy additional files to dist. */
+        copy({
+            targets: glob.sync('*')
```
carlo-lumapps, line 38: Not sure if we should put all files in the `dist` folder? Filtering the "source" files could make our library leaner. (It wouldn't make much difference in this case, I agree :p)

carlo-lumapps (approved):
Looks great!

I left one small comment, feel free to ignore it ^^

packages/lumx-icons/rollup.config.js
gcornut: I could list `*.scss`, `*.js`, `*.d.ts`, `package.json`, `*.md` (README/LICENCE). The result would be the same because we need to bundle all files but at least it would be more explicit than `*`

PR #869, opened 2022-08-01
fix(thumbnail): fix shift calculation when focusY is 0

hchevallier:
# General summary

When the container and the image size were equal, it caused a division by 0. 
It wasn't a big problem when the focusPoint is not equal to 0, because the result of a `x/0 === Infinity` if x !== 0, so it was translated to 100%. But `0/0 === NaN`, so when the focusPoint is 0, the computed CSS was invalid, so the focus point was not applied. 

Now we just early return a 0 if `containerSize === imageSize`. 
This PR also adds some tests on this function.
 

# Screenshots


gcornut (changes requested):
Logs should be removed

gcornut (approved):

gcornut:
Test OK: focus point works as expected on SB story :heavy_check_mark:

PR #871, opened 2022-08-02
fix(thumbnail): base shift calculation on scaledsize instead of imagesize

hchevallier:
# General summary

# Screenshots


gcornut (approved):

gcornut:
Test OK: tested on lumapps web with alpha release 2.2.23-alpha-thumbnail-focus-point.1

PR #875, opened 2022-09-13
fix(slideshow): improve accessibility

Mcourt:
# General summary

* Fixed keyboard navigation and screen reader compatibility for slideshows
  * Non visible slides now can't be navigated to using the keyboard.
  * Non visible slides are hidden from screen readers.
  * When slides become visible, the focus is restored and so is their screen reader visibility.

* Added roving tab index to controls
* Slides grouped together are now wrapped inside individual divs.
* Added roles and attributes to improve a11y complience.

# Screenshots
slideshow-focus-management


Mcourt:
Converting to draft as I am working on other improvements that may interfer.

gcornut (approved):

gcornut:
Test
- Slideshow focus order ✔️ 
- Slideshow hidden slides are removed from focus order ✔️ 
- Slideshow slide shortcut buttons  implement roving tab index ✔️ 

Non regression test
- Dialog focus trap works like before ✔️ 
- Slideshow looks and work as expected ✔️

PR #877, opened 2022-09-14
feat(heading): create heading component

Mcourt:
CF-365

# General summary

Create a `Heading` component to manage html headings (h1 / h2 / h3 / h4 / h5 / h6).
The component extends the `Text` component for styling but only accepts `HeadingElement` as `as` prop.

It also comes with a `HeadingLevelProvider` to allow the `HeadingComponent` to automatically know what level to use by default. The correct typography based on this level will also be applied.

Both the level and the typography can be overridden if needed.

# Screenshots


juanigalan91 (approved):

gcornut:
Test OK ✔️

PR #885, opened 2022-10-12
chore(storybook): improve stories `today` param handling

gcornut:
# General summary

When snapshoting stories in chromatic we need to inject the `today` Date so that it's always the same.

Previously this was injected in stories args but it can cause problems for stories that just sread the `args` as props. 

To improve the situation, the `today` value will be injected in the storybook `context.parameters`.

**Non regression tests**
- Check date picker [SB stories show the same default date as before](https://5fbfb1d508c0520021560f10-apdillnpmd.chromatic.com/?path=/story/lumx-components-date-picker-datepickerfield--simple)
- Check date picker demos (on the demo site) show the same default date as before
- Check chromatic snapshots still show the same date

### DIFF
--- a/packages/lumx-react/src/stories/generated/Heading/Demos.stories.tsx
+++ b/packages/lumx-react/src/stories/generated/Heading/Demos.stories.tsx
@@ -1,0 +1,6 @@
+/**
+ * File generated when storybook is started. Do not edit directly!
+ */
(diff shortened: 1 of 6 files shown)


juanigalan91 (approved):

gcornut:
Rebased

PR #886, opened 2022-10-12
feat(inline-list): create InlineList component

gcornut:
# General summary

A component to display inline list of elements separated by dots.

StoryBook: https://5fbfb1d508c0520021560f10-gtrtbxgvqy.chromatic.com/?path=/story/lumx-components-inline-list-inlinelist--mixed-no-wrap-and-truncate

# Screenshots

inline-list-doc

### DIFF
--- a/packages/lumx-core/src/scss/components/inline-list/_index.scss
+++ b/packages/lumx-core/src/scss/components/inline-list/_index.scss
@@ -1,0 +1,15 @@
+@use "sass:list";
+
+/* ==========================================================================
+   Inline list
+   ========================================================================== */
+
+.#{$lumx-base-prefix}-inline-list {
+    display: flex;
+    align-items: center;
+    list-style: none;
+
+    &__item {
+        display: contents;
+    }
+}
--- a/packages/lumx-react/src/components/inline-list/InlineList.tsx
+++ b/packages/lumx-react/src/components/inline-list/InlineList.tsx
@@ -1,0 +1,80 @@
+import React, { Children, forwardRef, isValidElement } from 'react';
+
+import classNames from 'classnames';
+
+import { ColorPalette, ColorVariant, Typography } from '@lumx/react';
+import { Comp, GenericProps } from '@lumx/react/utils/type';
+import { getFontColorClassName, getRootClassName, getTypographyClassName } from '@lumx/react/utils/className';
+
+/**
+ * Defines the props of the component.
+ */
+export interface InlineListProps extends GenericProps {
+    /**
+     * Text color.
+     */
+    color?: ColorPalette;
+    /**
+     * Lightened or darkened variant of the selected color.
+     */
+    colorVariant?: ColorVariant;
+    /**
+     * Typography variant.
+     */
+    typography?: Typography;
+}
+
+/**
+ * Component display name.
+ */
+const COMPONENT_NAME = 'InlineList';
+
+/**
+ * Component default class name and class prefix.
+ */
+const CLASSNAME = getRootClassName(COMPONENT_NAME);
+
+/**
+ * Component default props.
+ */
+const DEFAULT_PROPS = {} as const;
+
+/**
+ * InlineList component.
+ *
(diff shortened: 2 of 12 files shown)


packages/lumx-core/src/scss/components/inline-list/_index.scss
```diff
@@ -0,0 +1,15 @@
...
+   Inline list
+   ========================================================================== */
+
+.#{$lumx-base-prefix}-inline-list {
+    display: flex;
+    align-items: center;
+    list-style: none;
+
+    &__item {
+        display: contents;
```
carlo-lumapps, line 13: Never used if before ^^

Are you aware of this notice from MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents

packages/lumx-react/src/components/inline-list/InlineList.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const fontColorClassName = color && getFontColorClassName(color, colorVariant);
+    const typographyClassName = typography && getTypographyClassName(typography);
+    return (
+        // eslint-disable-next-line jsx-a11y/no-redundant-roles
+        <ul
+            {...forwardedProps}
+            ref={ref as any}
+            className={classNames(className, CLASSNAME, fontColorClassName, typographyClassName)}
+            // Lists with removed bullet style can lose their a11y list role on some browsers
+            role="list"
```
carlo-lumapps, line 60: Just a question: Why are we using the `role="list"` on a `ul` tag in this case?

carlo-lumapps (approved):
Cool ! 

I left two small questions but I'm approving it 👍

packages/lumx-core/src/scss/components/inline-list/_index.scss
gcornut: It's the only way I found to not break text overflow

packages/lumx-react/src/components/inline-list/InlineList.tsx
```diff
@@ -0,0 +1,80 @@
...
+        >
+            {Children.toArray(children).map((child, index) => {
+                const key = (isValidElement(child) && child.key) || index;
+                return (
+                    <li key={key} className={`${CLASSNAME}__item`}>
+                        {index !== 0 && (
+                            <span className={`${CLASSNAME}__item-separator`} aria-hidden="true">
+                                {'\u00A0•\u00A0'}
+                            </span>
+                        )}
```
carlo-lumapps, line 70: Could have been done with CSS only maybe ? 

```scss
.#{$lumx-base-prefix}-inline-list {
    &__item:nth-child(n + 2)::before {
        content: '\u00A0•\u00A0';
    }
}
```

@see https://css-tricks.com/examples/nth-child-tester/

@charles-lumapps FYI

gcornut: I did that first but this will be read by screen readers.

gcornut: Their is a way to put an empty alternative text on CSS content but not all browsers support it

packages/lumx-react/src/components/inline-list/InlineList.tsx
gcornut: I've written a comment above. Some browsers remove the list role when you remove list-style in CSS. So here I put it back in HTML to make sure it's there

packages/lumx-react/src/components/inline-list/InlineList.tsx
carlo-lumapps: Yes, I just saw it :) Interesting!

packages/lumx-react/src/components/inline-list/InlineList.tsx
carlo-lumapps: Thanks a lot 👍

PR #891, opened 2022-10-18
fix(inline-list): add listitem role to items

Mcourt:
# General summary

Since each element of the `InlineList` component has a `display: contents` css attribute, it removes the semantic value of these items.
Added back a `listitem` role on each items to reset the semantics.

# Screenshots


gcornut (approved):

gcornut:
Test OK
- stories showing list/listitem roles :heavy_check_mark: 
- demos showing text overflow properly :heavy_check_mark:

PR #914, opened 2022-11-14
chore(dialog): rework unmount to use CSS transition instead of timeout

gcornut:
# General summary

Use CSS opacity transition end instead of timeout to unmount the component (dialog, lightbox and notification).


gcornut:
Breaks when activating "prefers-reduced-motion" :x:

gcornut:
Should be fixed now

juanigalan91 (approved):

gcornut:
Test OK
- Storybook stories with dialog, notification and lightbox all work as expected ✔️ 
   => components unmount after the closing transition or immediately if `prefers-reduced-motion` is activated
- Lumapps web RTL tests all pass after some small fixes ✔️

PR #919, opened 2022-11-22
Fix/DIS-1053: not allowed cursor

radiuso:
DIS-1053

# General summary

As described in the mockups 
https://www.figma.com/file/pmsTWAYDud7j4PIKpXNmSS/Segments?node-id=1612%3A46301&t=t73FCW98y8TgkO4n-0

Disabled button should have a not-allowed cursor

Impact 
* Button
* Checkbox
* Chip
* link
* list
* radio-button
* select
* switch
* text-field

# Screenshots
Screenshot 2022-11-22 at 15 05 06

### DIFF
--- a/packages/lumx-core/src/scss/components/button/_mixins.scss
+++ b/packages/lumx-core/src/scss/components/button/_mixins.scss
@@ -130,7 +130,7 @@
         }
     }
 
-    &:hover,
+    &:hover:enabled,
     &[class*="--is-hovered"] {
         @if $color == "primary" and $emphasis == lumx-base-const("emphasis", "HIGH") {
             color: var(--lumx-button-#{$emphasis}-state-hover-#{$theme}-color);
@@ -156,7 +156,7 @@
         }
     }
 
-    &:active,
+    &:active:enabled,
     &[class*="--is-active"] {
         @if $color == "primary" and $emphasis == lumx-base-const("emphasis", "HIGH") {
             color: var(--lumx-button-#{$emphasis}-state-active-#{$theme}-color);
--- a/packages/lumx-core/src/scss/components/button/_mixins.scss
+++ b/packages/lumx-core/src/scss/components/button/_mixins.scss
@@ -130,7 +130,7 @@
         }
     }
 
-    &:hover:enabled,
+    &:hover:not(:disabled),
     &[class*="--is-hovered"] {
         @if $color == "primary" and $emphasis == lumx-base-const("emphasis", "HIGH") {
             color: var(--lumx-button-#{$emphasis}-state-hover-#{$theme}-color);
--- a/packages/lumx-core/src/scss/components/chip/_index.scss
+++ b/packages/lumx-core/src/scss/components/chip/_index.scss
@@ -111,7 +111,7 @@
     &.#{$lumx-base-prefix}-chip--color-light {
         @include lumx-state-as-selected(lumx-base-const("state", "DEFAULT"), lumx-base-const("theme", "DARK"));
 
-        &.#{$lumx-base-prefix}-chip--is-clickable:hover:enabled {
+        &.#{$lumx-base-prefix}-chip--is-clickable:hover:not(:disabled) {
             @include lumx-state-as-selected(lumx-base-const("state", "HOVER"), lumx-base-const("theme", "DARK"));
         }
 
--- a/packages/lumx-core/src/scss/components/chip/_index.scss
+++ b/packages/lumx-core/src/scss/components/chip/_index.scss
@@ -70,7 +70,7 @@
     .#{$lumx-base-prefix}-chip--is-unselected.#{$lumx-base-prefix}-chip--color-#{$key} {
         @include lumx-state(lumx-base-const("state", "DEFAULT"), lumx-base-const("emphasis", "MEDIUM"), $key);
 
-        &.#{$lumx-base-prefix}-chip--is-clickable:hover {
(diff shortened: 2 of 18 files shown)


packages/lumx-core/src/scss/core/state/_mixins.scss
```diff
@@ -107,7 +107,7 @@
 }
 
 @mixin lumx-state-disabled-input() {
-    pointer-events: none;
+    cursor: not-allowed;
```
gcornut, line 110: `pointer-events` must be kept otherwise you'll enable mouse event on all disabled elements that use this mixin (Some of those elements might not have a native `disabled` attribute that also disables all interactions).
 
```suggestion
    cursor: not-allowed;
    pointer-events: none;
```

gcornut: Hmm 🤔. So we have 2 solution

1. Remove pointer-events and make sure we don't have regression on:
    packages/lumx-core/src/scss/components/button
    packages/lumx-core/src/scss/components/checkbox
    packages/lumx-core/src/scss/components/chip
    packages/lumx-core/src/scss/components/link
    packages/lumx-core/src/scss/components/list
    packages/lumx-core/src/scss/components/radio-button
    packages/lumx-core/src/scss/components/select
    packages/lumx-core/src/scss/components/switch
    packages/lumx-core/src/scss/components/text-field
    And in `lumapps-web`.
    I tested on buttons, it looks like removing `pointer-event` will make the `hover` state appear in `disabled` were this was not present before 🤷 
2. Apply the cursor only on buttons like you said (but designers will probably ask for this cursor on all disabled components)

packages/lumx-core/src/scss/components/button/_mixins.scss
```diff
@@ -130,7 +130,7 @@
         }
     }
 
-    &:hover,
+    &:hover:enabled,
```
gcornut, line 133: When using the Button component with an `href` (instead of `onClick`) then we use `<a>` instead of `<button>` and it looks like it does not work with the `:enabled` CSS pseudo class. 
In the story "With Href", I don't see the hover state even though the Button is not disabled :S

packages/lumx-core/src/scss/components/chip/_index.scss
```diff
@@ -111,7 +111,7 @@
     &.#{$lumx-base-prefix}-chip--color-light {
         @include lumx-state-as-selected(lumx-base-const("state", "DEFAULT"), lumx-base-const("theme", "DARK"));
 
-        &.#{$lumx-base-prefix}-chip--is-clickable:hover {
+        &.#{$lumx-base-prefix}-chip--is-clickable:hover:enabled {
```
gcornut, line 114: Same here, `:enabled` does not work with `<a>` used in Chip. 
This is testable in the "Clickable" story, when activating the dark background, the hover state does not work

packages/lumx-core/src/scss/components/text-field/_index.scss
```diff
@@ -187,7 +188,7 @@
 
 // Hover state
 .#{$lumx-base-prefix}-text-field--theme-light {
-    .#{$lumx-base-prefix}-text-field__wrapper:hover {
+    .#{$lumx-base-prefix}-text-field__wrapper:hover:enabled {
```
gcornut, line 191: Same here `:enabled` does not work in `<div>`

packages/lumx-core/src/scss/components/text-field/_index.scss
```diff
@@ -208,7 +209,7 @@
 }
 
 .#{$lumx-base-prefix}-text-field--theme-dark {
-    .#{$lumx-base-prefix}-text-field__wrapper:hover {
+    .#{$lumx-base-prefix}-text-field__wrapper:hover:enabled {
```
gcornut, line 212: Same here `:enabled` does not work in `<div>`

gcornut (changes requested):

packages/lumx-core/src/scss/components/text-field/_index.scss
radiuso: Should be fixed

gcornut (approved):

Mcourt:
From our test session on storybook: 

**Button**
- [X] Click doesn't go through when disabled 
- [X] Mouse style (click / hover) doesn't apply disabled

**Checkbox**
- [X] Click doesn't go through when disabled 
- [X] Mouse style (click / hover) doesn't apply disabled

**Chip**
- [X] Click doesn't go through when disabled 
- [X] Mouse style (click / hover) doesn't apply disabled

**Link**
- [X] Click doesn't go through when disabled 
- [X] Mouse style (click / hover) doesn't apply disabled

**List**
- [X] Click doesn't go through when disabled 
- [ ] Mouse style (click / hover) doesn't apply disabled :x: 

**Radio-button**
- [X] Click doesn't go through when disabled 
- [X] Mouse style (click / hover) doesn't apply disabled

**Select**
- [ ] Click doesn't go through when disabled :x: 
- [ ] Mouse style (click / hover) doesn't apply disabled :x:
Hover style displayed on input when hovered => should not.

**Switch**
- [x] Click doesn't go through when disabled 
- [x] Mouse style (click / hover) doesn't apply disabled

**Text-field**
- [x] Click doesn't go through when disabled 
- [x] Mouse style (click / hover) doesn't apply disabled

gcornut:
Closing this PR because we'll rework the disabled state to also make disabled component focusable for better a11y

PR #925, opened 2022-11-30
feat(popover-dialog): Create PopoverDialog component

Mcourt:
# General summary

Create a new PopoverDialog component that has the same design and API as a `Popover` component but uses the `dialog` accessible pattern.

# Screenshots

**Story**
popover-dialog

**Documentation**
image


packages/lumx-react/src/utils/type.ts
```diff
@@ -91,3 +91,35 @@ export const isComponentType = (type: ReactElement['type']) => (node: ReactNode)
  * (excluding `NaN` as it can't be distinguished from `number`)
  */
 export type Falsy = false | undefined | null | 0 | '';
+
+/**
+ * Allow at least one of aria-label or arial-labelledby props to be set.
+ */
+export type WithAriaLabelOrLabelledBy =
```
gcornut, line 98: We should name it more like `HasTheme`
```suggestion
export type HasAriaLabelOrLabelledBy =
```

packages/lumx-react/src/utils/type.ts
```diff
@@ -91,3 +91,35 @@ export const isComponentType = (type: ReactElement['type']) => (node: ReactNode)
...
+ * or one of aria-label or arial-labelledby prop.
+ */
+export type WithLabelPropOrAriaLabel<T = string | undefined> = T extends string
+    ? {
+          /**
+           * The label of the dialog.
+           * Must define the purpose of the dialog.
+           * Alternatively, you can use `aria-label` directly or target an existing element as title using `aria-labelledby`
+           */
+          label: T;
```
gcornut, line 119: It's probably best to use aria-label directly since the label is not used in DOM or elsewhere (like the IconButton)

packages/lumx-react/src/components/popover-dialog/PopoverDialog.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+        // Tab to next button
+        await userEvent.tab();
+
+        // As there is no more button, focus should loop back to first button.
+        expect(dialogButtons[0]).toHaveFocus();
+
+        // Close the popover
+        await userEvent.keyboard('{escape}');
+
+        expect(screen.queryByRole('dialog', { name: label })).not.toBeInTheDocument();
```
gcornut, line 60: Can we also check that the focus goes back to the anchor ?

gcornut (changes requested):

packages/lumx-react/src/utils/type.ts
Mcourt: Sure thing ! Changed :)

packages/lumx-react/src/utils/type.ts
Mcourt: Got it, removed this type and kept only the other!

packages/lumx-react/src/components/popover-dialog/PopoverDialog.test.tsx
Mcourt: Good call! It's done

packages/site-demo/content/product/components/popover-dialog/index.mdx
```diff
@@ -0,0 +1,26 @@
...
+This component should be used when popovers contains complex content with multiple interactive elements.
+
+As this component uses the `dialog` accessible pattern, a few specific behaviors are applied:
+
+- When the popover is opened, focus will be set on either the element set using the `focusElement` prop or the first focusable element.
+Once the focus is set within the popover, it will be trapped and stay within the popover until it is closed.
+
+- A `label` must be defined using either `label`, `aria-label` or `aria-labelledby` props.
+Setting the `label` prop will apply an `aria-label` to the dialog but `aria-labelledby` can also be used to target
+a visible element as popover title.
```
gcornut, line 18: Some minor fix and also maybe we could recommend `aria-labelledby` more than `aria-label` ?
```suggestion
- A label must be defined using either `aria-label` or `aria-labelledby` props.
  Setting `aria-labelledby` is recommended as it target a visible label rather than adding an invisible `aria-label` on the dialog.
```

gcornut (changes requested):

packages/site-demo/content/product/components/popover-dialog/index.mdx
Mcourt: Sure. I've also inverted the types priorities so that the first error to popup is for the missing `aria-labelledby` prop. :)

gcornut (approved):

gcornut:
Test OK
- SB stories work as expected ✔️ 
- doc page (+demo) work as expected ✔️

PR #929, opened 2022-12-05
fix(button): fix color props allowing string not in colorpalette

Fafnor:
[CF-407](https://lumapps.atlassian.net/browse/CF-407)

# General summary
Change the type of the color prop of the DS components to accept only ColorPalette values (instead of accepting string too)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Progress: split into `ProgressCircular` and `ProgressLinear` components. `Progress` is now deprecated.
 -   ProgressCircular: add `size` variants.
 
+### Changed
+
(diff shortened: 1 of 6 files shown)


carlo-lumapps (approved):

packages/lumx-react/src/components/index.ts
```diff
@@ -34,7 +34,6 @@ export const ColorPalette = {
     grey: 'grey',
 } as const;
 export type ColorPalette = ValueOf<typeof ColorPalette>;
-export type Color = ColorPalette | string;
```
gcornut, line 37: I think we should keep the `Color` type as an alias to the `ColorPalette` so that users do not have to replace it from their import.

gcornut (changes requested):

gcornut:
Rebased & improved changelog message

gcornut (approved):

gcornut:
On hold to test type error inpact in lumapps-web 
(also v3.0.7 will soon be merged to master and this PR will need to be rebased and changelog will need to be updated to make sure the change in this PR appear in the right section)

gcornut:
Test OK ✔️

PR #937, opened 2022-12-08
chore(progress): add size to circular variant

radiuso:
# General summary

Add size to the Circular Progress.
All size available 'xxs' | 'xs' | 's' | 'm'

# Screenshots


packages/lumx-core/src/scss/components/progress/_index.scss
```diff
@@ -1,21 +1,46 @@
...
+            height: map.get($lumx-sizes, $key);
+
+            #{$self}__svg {
+                position: absolute;
+                top: 0;
+                right: 0;
+                bottom: 0;
+                left: 0;
+                display: var(--lumx-material-progress-rotate);
+                width: map.get($lumx-sizes, $key);
```
matmkian, line 36: Keep the global definition of svg and only define in the map things related to the size to avoid code duplication

matmkian (changes requested):

packages/lumx-react/src/components/progress/Progress.tsx
```diff
@@ -12,15 +12,29 @@ import { getRootClassName, handleBasicClasses } from '@lumx/react/utils/classNam
...
+    /** Progress variant. */
+    variant?: 'linear';
+}
+
+export type ProgressProps =
+    | ProgressCircularProps
+    | (ProgressLinearProps & {
+          /** Progress variant. */
+          variant?: ProgressVariant;
+      });
```
gcornut, line 36: We need that:
```suggestion
    variant: 'linear';
}

export type ProgressProps = ProgressCircularProps | ProgressLinearProps;
```
I order to have `<Progress size="foo" />` return an error since the default variant is 'circular' and thus the size must use the correct format

gcornut (changes requested):

gcornut (approved):

gcornut:
Test OK 
- non regression (SB) :heavy_check_mark: 
- doc page :heavy_check_mark: 
- size variants (SB) :heavy_check_mark:

matmkian (approved):

PR #938, opened 2022-12-08
feat(grid-column): create new GridColumn component

armasson:
# General summary

Add a new component : `GridColumn` that display children in a customised grid, with a responsive number of column based on child min width. The number of column shown will reduce when there is not enough space available to show each item with the specified min-width.

# Screenshots


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 -   Checkbox: add `disabled` attribute to the `input` element
+-   New component in React `GridColumn`
```
gcornut, line 15: Need to move this under "Unreleased". It's probably because you rebased once 3.0.7 was merged and git does not see this as a conflict

packages/lumx-react/src/components/grid-column/GridColumn.tsx
```diff
@@ -0,0 +1,90 @@
...
+                    className,
+                    handleBasicClasses({
+                        prefix: CLASSNAME,
+                    }),
+                )}
+                style={{
+                    ...style,
+                    ['--lumx-grid-column-item-min-width' as any]: isInteger(itemMinWidth) && `${itemMinWidth}px`,
+                    ['--lumx-grid-column-columns' as any]: maxColumns,
+                    ['--lumx-grid-column-gap' as any]: gap && `var(--lumx-spacing-unit-${gap})`,
```
gcornut, line 80: I rather have the gap as a classique modifier class like in flexbox https://github.com/lumapps/design-system/blob/05042e921207cafffc4e0b5a82b741f76efd06b3/packages/lumx-react/src/components/flex-box/FlexBox.tsx#L85
Instead of a CSS variable since it's a finite list of values

packages/site-demo/content/product/components/grid-column/index.mdx
```diff
@@ -0,0 +1,27 @@
+import { GridColumn } from '@lumx/react';
+import { Placeholder } from '@lumx/demo/components/content/Placeholder';
+
+# GridColumn
+
+**Display children in a customised grid, with a responsive number of column based on child min width**
+
+### Accessibility concerns
```
gcornut, line 8: This section is always at the end of component pages, just before the properties section

gcornut (changes requested):

packages/lumx-react/src/components/grid-column/GridColumn.tsx
armasson: I'll still need the CSS variable, for the `grid-template-column` calculation, since we take into account the gap between items.

gcornut:
Test OK (SB :heavy_check_mark: doc :heavy_check_mark: unit tests :heavy_check_mark:

PR #948, opened 2023-01-02
fix(tab): fix css to make sure disabled tab is not displayed as active

DamienGarnes:
[DS-109](https://lumapps.atlassian.net/browse/DS-109)

# General summary
When a tab is disabled, the style was applied only if it's selected. It's now fixed, even if the tab is not active.

# Screenshots
image


FlorentDinet:
Nice work, all good for me

gcornut (approved):

PR #958, opened 2023-01-24
chore(popover): refactor & fix RTL act() errors

gcornut:
- Extract all logic in a hook
- Extract constants in separate file
- Skip logic on jsdom env
- Add `as` prop


gcornut:
Test OK
Non regression testing on lumapps-web ✔️

Mcourt (approved):

PR #972, opened 2023-04-19
fix(select): fix select dropdown focus not being trapped

Fafnor:
# General summary
Fix select dropdown focus within a dialog not being trapped

This PR trap the focus on the Select dropdown like it is the case with a Dialog, you can test it with the story SelectWithinADialog https://www.chromatic.com/builds?appId=5fbfb1d508c0520021560f10


gcornut:
Test OK focus trap activated on select open

PR #974, opened 2023-04-24
fix(popover): fix scroll on open/close

gcornut:
- fix(focus-trap): remove unnecessary scroll on focus trap activation
- fix(popover): remove unnecessary scroll on close


Mcourt (approved):

gcornut:
Test OK, no regression seen on the product

PR #977, opened 2023-04-28
chore(core): remove deprecated sass tilde imports

gcornut:
# General summary

1. Remove `~` import in SCSS that are now deprecated (module import work fine without it)
2. Lint all SCSS files


gcornut:
I compared the produced CSS on this PR vs on master and the file are literally identical (same sha1 hash) 👍

Mcourt (approved):

PR #982, opened 2023-05-23
fix(iconbutton): remove ariadescribedby attribute when not needed

Fafnor:
# General summary
Screen reader was reading two time the icon button label on focus because of the label being present in aria-label and on aria-describedby (which read the tooltip label). 

This PR aim to remove the aria-describedby attribute on icon button when the aria-label and the tooltip label are the same

Tests steps:
- Go to the tooltip component and check that when you focus the button you have an aria-describedby attribute
- Go to the iconButton Tooltip component and check that when you focus the button you don't see the aria-describedby attribute

### DIFF
--- a/packages/lumx-react/src/components/button/IconButton.tsx
+++ b/packages/lumx-react/src/components/button/IconButton.tsx
@@ -61,7 +61,13 @@ export const IconButton: Comp<IconButtonProps, HTMLButtonElement> = forwardRef((
 
     return (
         <Tooltip label={hideTooltip ? '' : label} {...tooltipProps}>
-            <ButtonRoot ref={ref} {...{ emphasis, size, theme, ...forwardedProps }} aria-label={label} variant="icon">
+            <ButtonRoot
+                ref={ref}
+                {...{ emphasis, size, theme, ...forwardedProps }}
+                aria-label={label}
+                variant="icon"
+                aria-describedby={tooltipProps?.label && tooltipProps?.label !== label ? null : undefined}
+            >
                 {image ? (
                     <img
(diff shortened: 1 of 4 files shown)


packages/lumx-react/src/components/button/IconButton.tsx
```diff
@@ -61,7 +61,13 @@ export const IconButton: Comp<IconButtonProps, HTMLButtonElement> = forwardRef((
 
     return (
         <Tooltip label={hideTooltip ? '' : label} {...tooltipProps}>
-            <ButtonRoot ref={ref} {...{ emphasis, size, theme, ...forwardedProps }} aria-label={label} variant="icon">
+            <ButtonRoot
+                ref={ref}
+                {...{ emphasis, size, theme, ...forwardedProps }}
+                aria-label={label}
+                variant="icon"
+                aria-describedby={tooltipProps?.label && tooltipProps?.label !== label ? null : undefined}
```
Mcourt, line 69: I'm not sure why we need to do this? 
Why the distinction between null and undefined? Is it because it will be overridden afterwards and "null" completely removes it while undefined doesn't?
Could you add a comment explaining what's happening?

Fafnor: `null` was not useful here, I removed it and added a comment

Mcourt (approved):

Mcourt:
TEST: ok

:heavy_check_mark: Tooltip is still correctly displayed on Tooltip and IconButton stories
:heavy_check_mark: Tooltip text on IconButtons are only read once using NVDA
:heavy_check_mark: Tooltip text is correctly read when different than the icon button's label.

PR #985, opened 2023-06-08
chore(storybook): configure storybook test runner

gcornut:
# General summary

Configure storybook test runner

- Dev use case: run storybook & run test:storybook --watch to code & test component in storybook interactions
- CI use case: run stories storybook interactions tests (with fallback to smoke test)


gcornut:
Needs rework

PR #986, opened 2023-06-08
fix(uploader): improve a11y & doc ARCH-59

gcornut:
- fix(uploader): use true button to improve accessibility
- feat(uploader): add native file input handler
- docs(uploader): add accessibility concerns section

Use a true `<button>` button by default to improve accessibility. 
Use `<input type=file>` and `<label>` if `fileInputProps` is provided to handle a native file picker and to support dropping a file onto the uploader.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   TableCell: fix clickable & sortable header cell accessibility
--   Uploader: use true button element for better accessibility
+-   Uploader: use button element by default
+-   Uploader: use label and input file elements when providing `fileInputProps`
 
 ## [3.3.1][] - 2023-06-12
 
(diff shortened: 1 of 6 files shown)


Mcourt (approved):

Mcourt (changes requested):
Wait actually, I'm no too sure about this.
If we see "Uploaders" as buttons that opens the native file picker, I'm not so sure they should be "buttons".
There might be some places where it should be a "label" with a file input linked to it.
Like we could see on these resources:
[Medium - Accessible File Uploads
](https://medium.com/@christophergoodwin.me/accessible-file-uploads-9ecacac17276)
[Accessibility.com - A styled accessible file upload
](https://accessabilly.com/a-styled-accessible-file-upload/)

If this is not the intended use, we could probably add more context in the doc section?
We should at least be able to use it like this somehow.

Mcourt (approved):
After some further research, I have found some resource that seem to show that this solution can be acceptable.
Approving and it can always be enhanced later.

gcornut:
Deployed StoryBook: https://5fbfb1d508c0520021560f10-ybfgjlvwgt.chromatic.com/?path=/story/lumx-components-alert-dialog-alertdialog--default-kind

PR #992, opened 2023-06-14
chore(table): improve a11y & doc ARCH-53

gcornut:
- fix(table-cell): fix clickable & sortable table cell accessibility
- docs(table): document accessibility concerns


Mcourt (approved):
Nice!

PR #994, opened 2023-06-14
docs(notification): document accessibility concerns ARCH-42

gcornut:


packages/site-demo/content/product/components/notification/index.mdx
```diff
@@ -4,6 +4,12 @@
 
 <DemoBlock orientation="horizontal" demo="default" hasPlayButton />
 
+### Accessibility concerns
+
+Notifications use the ARIA `alert` role which will announce itself when it appears.
+
+Clickable notifications (or containing clickable action button) are not easily accessible with the keyboard. Consider moving the focus to the clickable elements and restore the focus to the previous place when the notification is closed.
```
Mcourt, line 11: Should this be built into the notification when an action is set? Maybe that should be the default behavior.

Maybe in a later PR?

gcornut: I don't see how we would do that for now and does not look like we actually use this feature ^^'

So yeah if it pops up as an issue some day we could look into solving this. Or we could deprecate this entirely

Mcourt (approved):

PR #996, opened 2023-06-19
fix(autocomplete): focus text field on clear button clicked

Mcourt:
# General summary

Since the autocomplete passes a `mergeRefs` function as inputRef and that the TextField didn't manage this case, the focus was not restored to the input field on clear.

Fixed by adding another "ref" locally to the TextField to manage the focus.

# Screenshots
autocomplete-focus-on-clear

StoryBook: https://fc535bf8--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=289)) **⚠️ Outdated commit** **⚠️ Outdated commit**


gcornut:
Test OK, focus is restored to the text field when clicking (or activating with key press) the clear button

PR #1022, opened 2023-10-17
feat(list-item): add focus visible class to use with virtual focus

gcornut:
Move the outline style on key nav to make it also available via a class `.lumx-list-item__link--focus-visible` to be used in virtual focus (like in select/autocomplete, where the list item __appears__ focused but the input is the element focused)

TODO:
- [ ] Use in autocomplete/select ?
- [ ] Add as prop to ListItem ?
- [ ] Validate


gcornut:
Not needed right now

PR #1024, opened 2023-11-06
chore(typography): remove default line height on custom title typography

npankov:
Fix: https://lumapps.atlassian.net/browse/SUP-4003 

Remove the default line height on custom title typography to avoid miss match with font size when customizing CSS variables

TODO:
- [x] Test in demo site
- [ ] Test in prod env
- [ ] Rebase and resolve conflict on CHANGELOG.md file


packages/lumx-core/style-dictionary/properties/core/typography.json
```diff
@@ -59,7 +59,7 @@
             "title1": {
                 "font-size": { "value": "40px" },
                 "font-weight": { "value": "var(--lumx-typography-font-weight-bold)" },
-                "line-height": { "value": "50px" }
+                "line-height": { "value": "1.5" }
```
matmkian, line 62: Line height is 1.5 by default -> just remove all typography line heights and it will inherit from 1.5 line height from the DS

matmkian (changes requested):

matmkian:
> Personnally I would keep the line-height option for custom titles, as we used it for some client's CSS and if it disappears it would have some impact.

@SophieChristophe the variable will still be there but no default value so it inherits from 1.5 global line height. So we have to remove the default values but let define line heights if wanted.

SophieChristophe:
> > Personnally I would keep the line-height option for custom titles, as we used it for some client's CSS and if it disappears it would have some impact.
> 
> @SophieChristophe the variable will still be there but no default value so it inherits from 1.5 global line height. So we have to remove the default values but let define line heights if wanted.

That's OK for me, just want to make sure the available customizations don't disappear

matmkian (approved):

gcornut (changes requested):
I think it's a bit risky to remove variables, if anyone uses the line height SCSS variables in their code, they will get an "undefined variable" error. 
I think the solution would be to set the line height variables to `null`. Null variables in SCSS are simply ignored, just what we want. 
But I need to update the CSS variable generation to ignore null values (not a valid value in CSS)

matmkian:
> I think it's a bit risky to remove variables, if anyone uses the line height SCSS variables in their code, they will get an "undefined variable" error. I think the solution would be to set the line height variables to `null`. Null variables in SCSS are simply ignored, just what we want. But I need to update the CSS variable generation to ignore null values (not a valid value in CSS)

+1 for null value rather than removing it

gcornut (approved):

npankov:
## Tests OK ✅ 
✅ scroll bar is not displayed on the custom title windows + chrome and other's
✅ line-height inherit from 1.5 line height from the DS

gcornut:
Rebased and fixed changelog conflict

PR #1028, opened 2023-11-16
chore(lumx-icons): override twitter icon

gcornut:
# General summary

1. Creating a script to override and alias icons in both TS and SCSS/Font
    - `override/override-icons`: contains SVG icons to override MDI
    - `override/alias-icons.js`: config contains mapping from icon name to aliases
    - `override/generate.js`: script that generate the TS & Font/SCSS overrides and then TS & Font/SCSS aliases

We'll use aliases to keep retro-compatibility when changing version or for brand renaming (ex: twitter). Overrides will help us update or restore brand logo because MDI is deprecating them.

2. Override Twitter logo with X logo (and add name alias)

# Testing

1. Publishing an alpha prerelease `alpha-icon-override-twitter` https://github.com/lumapps/design-system/actions/runs/6956038315/job/18962918745
2. Test in another project using both JS & font icons
     1. JS const `mdiTwitter` should still work but should have been swapped to the new X logo
     2. JS const `mdiXTwitter` should be the same as `mdiTwitter`
     3. css class `.mdi-twitter` should still work but should have been swapped to the new X logo
     4. css class `.mdi-x-twitter` should be the same as `.mdi-twitter`


packages/lumx-icons/override/generate.js
```diff
@@ -0,0 +1,172 @@
...
+    .then(() =>
+        Promise.all([
+            generateMergedFont(),
+            generateTSOverride(),
+            generateSCSSOverride(),
+            generateTSAliases(),
+            generateSCSSAliases(),
+        ]),
+    )
+    .then(cleanUp);
```
juanigalan91, line 172: Maybe a couple of comments here, explaining what each functions does will make it easier to understand what the script does. Each function is a bit cryptic, they do a lot of file reading, parsing etc, that is not easy to comprehend by just reading them.

packages/lumx-icons/override/generate.js
```diff
@@ -0,0 +1,172 @@
...
+ */
+async function generateTSOverride() {
+    const icons = await readJSON(path.resolve(GENERATED_TMP_DIR, `${OVERRIDE_FONT_NAME}.json`));
+    const out = [`export * from '@mdi/js';`];
+    for (const [icon, paths] of Object.entries(icons)) {
+        if (paths.length !== 1) throw new Error('Only single path icon are supported');
+        out.push(`export const ${camelCase(`mdi-${icon}`)} =\n    '${paths[0]}';`);
+    }
+    const outFile = path.resolve(GENERATED_DIR, `${OUT_OVERRIDE_NAME}.ts`);
+    await fs.writeFile(outFile, `${out.join('\n\n')}\n`);
```
juanigalan91, line 67: maybe some comments for each function would be interesting as well

packages/lumx-icons/rollup.config.js
```diff
@@ -44,8 +45,32 @@ export default {
...
-                .map(file => ({ src: path.join(__dirname, file), dest: DIST_PATH })),
+            targets: glob.sync('*(package.json|*.scss|*.md)').concat(glob.sync('override/generated/{*.scss,fonts}'))
+                .map(file => ({ src: path.join(__dirname, file), dest: path.join(DIST_PATH, path.dirname(file)) })),
         }),
     ],
 };
+
+
+// Bundle TS types in D.TS files
+const bundleType = {
```
juanigalan91, line 56: Is there a specific reason why we need types on the released version for our icons?

gcornut: Yes the mdi icon types are really trivial but when you import non-typed module in TS project it can produce errors or warnings (depending on the project config). We did had types before from mdi (and from the v4-to-v5 alias file), I just replaced manual type definition with automatic `.d.ts` bundling (we already use on `@lumx/react`)

juanigalan91 (approved):

gcornut:
Tested in another project using both JS & font icons
- JS const mdiTwitter should still work but should have been swapped to the new X logo ✅ 
- JS const mdiXTwitter should be the same as mdiTwitter ✅ 
- css class .mdi-twitter should still work but should have been swapped to the new X logo ✅ 
- css class .mdi-x-twitter should be the same as .mdi-twitter ✅

PR #1062, opened 2024-01-04
chore(design-tokens): add font-weight medium

armasson:
# General summary

The `medium` value was not present  for the `font-weight` typography in the design token.

# Screenshots


gcornut (approved):

gcornut:
No impact on existing code and change is trivial so we won't need to test this

PR #1063, opened 2024-01-05
chore(tooltip): fix escape key handler only when open

gcornut:
- chore(tooltip): make sure to only listen escape key on open
- chore(tooltip): add tests on close on escape

Not modifying the CHANGELOG because this fix is on a bug not yet releases

StoryBook: https://5991f7f4--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=323))

### DIFF
--- a/packages/lumx-react/src/components/tooltip/Tooltip.test.tsx
+++ b/packages/lumx-react/src/components/tooltip/Tooltip.test.tsx
@@ -1,13 +1,8 @@
 import React from 'react';
 
 import { Button } from '@lumx/react';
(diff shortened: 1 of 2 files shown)


carlo-lumapps (approved):

gcornut:
Test OK
- Tooltip closes on escape ✔️ 
- Tooltip does not block close on escape of other components (dialogs, popover, etc.) when closes ✔️

PR #1072, opened 2024-01-11
feat(popover): rework arrow style and add dark theme

gcornut:
https://lumapps.atlassian.net/browse/DSW-43

- feat(popover): rework arrow style
  Make arrow bigger and make the drop shadow fit better with the whole popover
  | before | after |
  | --- | --- |
  | image | image |
  
  Test on: https://257497c2--5fbfb1d508c0520021560f10.chromatic.com/?path=/story/lumx-components-popover-popover--simple&args=hasArrow:true

- feat(popover): add dark theme
  Add dark theme changing the popover and arrow background
  image

  Test on: https://257497c2--5fbfb1d508c0520021560f10.chromatic.com/?path=/story/lumx-components-popover-popover--dark-theme

StoryBook: https://257497c2--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=330))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Thumbnail: fix warnings on image overflow style
 
+### Changed
+
+-   Popover: reworked arrow style to be bigger and with a correct drop shadow
+
 ## [3.6.2][] - 2024-01-16
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


hchevallier (approved):

gcornut:
We had a small difference between the blur size we had before with box-shadow and the blur we now have in drop-shadow. It looks like the blur size in drop-shadow needs to be divided by two to match the blur we have in box-shadow

PR #1084, opened 2024-02-06
feat(notification): add `usePortal` prop

gcornut:
# General summary

Add `usePortal` prop on the notification component to make it render outside the default React portal


gcornut:
Releasing an alpha version https://github.com/lumapps/design-system/actions/runs/7802901510

hchevallier (approved):
LGTM

gcornut:
Added a fix to correctly forward the styles to the root div

PR #1088, opened 2024-02-13
fix(input-helper): replace span with p element to improve semantic

al3xbzh:
# General summary

For a11y, improve the semantic of the inputHelper component by using a semantic `<p>` tag instead of a `<span>`.

# Screenshots

No visual changes, only the markup changes.


gcornut (approved):

gcornut:
Test OK - no visual regression, the input helper span is indeed replaced by a p

PR #1092, opened 2024-02-20
fix(tooltip): avoid removing the anchor aria-describedby

gcornut:
# General summary
Avoid removing the anchor aria-describedby value:
- Concat with the tooltip id if the tooltip label is not the same as the anchor label
- Keep it as-is otherwise


hchevallier (approved):
LGTM 
Nice work!

PR #1094, opened 2024-02-21
chore: fix v3.6.4 regressions

gcornut:
- Revert "fix(tooltip): fix unstable anchor ref with memoize"
  Memoizing the tooltip anchor ref seems to break some complex tests

- fix(tooltip): do not add aria-describedby when closed
  Description should only be injected when opened

- refactor(tooltip): improve children element detection
  
- chore(utils): make isFocusVisible default to `true` on test env
  Better default return value in test env (jsdom does not support :focus-visible)

- fix(popover): unstable restore focus and changing anchor refs
  Fix restore focus on icon button (wrapped in tooltip) when popover dialog closes

- feat(autocomplete): forward `focusAnchorOnClose` prop
  Required in some use cases


gcornut:
Releasing an alpha version https://github.com/lumapps/design-system/actions/runs/7990020352...

gcornut:
Releasing an alpha version  https://github.com/lumapps/design-system/actions/runs/7990573462

hchevallier (approved):

gcornut:
Test OK ✔️ - tested in lumapps

PR #1098, opened 2024-02-27
fix(thumbnail): fix child DOM tag semantic

gcornut:
# General summary

https://lumapps.atlassian.net/browse/DSW-135

Use `span` instead of `div` in children thumbnail elements to avoid semantic error on clickable thumbnails (button).

**Test scenario**: non-regression visual tests on thumbnails
- Check clickable and non clickable thumbnails
- Check with or without aspect ratio
- Check image fallback

StoryBook: https://5fbfb1d508c0520021560f10-lwksiktedo.chromatic.com/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


hchevallier (approved):
LGTM

gcornut:
Test OK - no visual regression

PR #1100, opened 2024-03-08
chore(storybook): migrate to v8 with webpack and swc

gcornut:
# General summary

Migrating to Storybook v8 with webpack and SWC

Will require node 18+


gcornut:
Needs rework

PR #1107, opened 2024-03-21
feat(dialog): make dialog full size on mobile

matmkian:
# General summary

# Screenshots

### DIFF
--- a/packages/lumx-core/src/scss/components/dialog/_index.scss
+++ b/packages/lumx-core/src/scss/components/dialog/_index.scss
@@ -40,6 +40,10 @@
         background-color: lumx-color-variant("dark", "L1");
         animation: lumx-dialog-overlay-show $lumx-dialog-transition-duration;
 
+        @include mq($until: desktop) {
+            display: none;
+        }
+
         @media (prefers-reduced-motion: reduce) {
             animation: none;
         }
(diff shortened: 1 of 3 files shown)


packages/lumx-core/src/scss/components/dialog/_index.scss
```diff
@@ -51,22 +55,39 @@
...
+
+        #{$self}--is-hidden & {
+            @include mq($until: desktop) {
+                transform: translateY(100%);
+            }
+        }
+
+        @include mq($until: desktop) {
+            transition: transform 500ms cubic-bezier(0.23, 1, 0.32, 1);
+            animation: lumx-dialog-show-mobile 500ms cubic-bezier(0.23, 1, 0.32, 1);
```
hchevallier, line 67: Could we store the duration in a variable `$lumx-responsive-dialog-transition-duration`, it is useful to synchronize other animations with dialog opening/closing. (`$lumx-dialog-transition-duration` is used in Lumapps)

hchevallier (approved):
LGTM

FlorentDinet:
Do we have a mobile breakpoint to use instead of tablet?

I think its a bit too much for confirm dialog on iPad for example:
Capture d’écran 2024-03-25 à 14 27 14

FlorentDinet:
Ok for this change, we'll optimise later the breakpoint

PR #1110, opened 2024-03-22
feat(message): add a close button to the message component

hchevallier:
# General summary
https://lumapps.atlassian.net/browse/DSW-137

## Message
- Adds a new prop `closeButtonProps` to handle a new close button. Only the `info` kind with a background supports it. The button won't render for any other configuration. The label is required.

# Screenshots

image

StoryBook: https://86b3eb90--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=359)) **⚠️ Outdated commit**

### DIFF
--- a/packages/lumx-react/src/components/message/Message.stories.tsx
+++ b/packages/lumx-react/src/components/message/Message.stories.tsx
@@ -5,6 +5,7 @@ import { loremIpsum } from '@lumx/react/stories/utils/lorem';
 import { iconArgType } from '@lumx/react/stories/controls/icons';
 import { withCombinations } from '@lumx/react/stories/decorators/withCombinations';
 import { withUndefined } from '@lumx/react/stories/controls/withUndefined';
+import { withNestedProps } from '../../stories/decorators/withNestedProps';
 
 export default {
     title: 'LumX components/message/Message',
@@ -53,4 +54,19 @@ export const CustomIcon = {
     args: {
         icon: mdiDelete,
     },
+};
+
+/**
+ * With close button (has background and kind info)
+ */
+export const ClosableMessage = {
+    args: {
+        'closeButtonProps.label': 'Close',
+        hasBackground: true,
+        kind: 'info',
+    },
+    argTypes: {
+        'closeButtonProps.onClick': { action: true },
+    },
+    decorators: [withNestedProps()],
 };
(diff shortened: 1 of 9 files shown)


packages/lumx-react/src/components/message/Message.stories.tsx
```diff
@@ -54,3 +54,17 @@ export const CustomIcon = {
...
+ */
+export const ClosableMessage = {
+    args: {
+        closeButtonProps: {
+            onClose: () => console.log('test'),
+            label: 'Close',
+        },
+        hasBackground: true,
+        kind: 'info',
+    },
```
gcornut, line 69: Can we remove the console.log in favor of a SB action ?

Nested objects break when trying to combine `args` and `argTypes` but we have a `withNestedProps()` decorator that handles a custom `<prop>.<nestedprop>` notation to nest object 

```suggestion
    args: {
        'closeButtonProps.label': 'Close',
        hasBackground: true,
        kind: 'info',
    },
    argTypes: {
        'closeButtonProps.onClick':  { action: true },
    }
    decorators: [withNestedProps()]
```

packages/site-demo/content/product/components/message/react/closable.tsx
```diff
@@ -0,0 +1,11 @@
...
+import React from 'react';
+
+export const App = () => (
+    <Message kind={Kind.info} hasBackground closeButtonProps={{ label: 'Close', onClose: () => console.log('Close') }}>
+        <p>
+            Message text quisque tincidunt lobortis dui non auctor. Donec porta, ligula volutpat vehicula aliquet, dui
+            sapien tempus felis, sed cursus diam ante.
+        </p>
+    </Message>
+);
```
gcornut, line 11: I'm not sure we should demo like this. I think we should make it interactive and have it actually close 
But then maybe we need a button to reopen the message when it is closed 🤔

gcornut (changes requested):

packages/site-demo/content/product/components/message/react/closable.tsx
hchevallier: That seems to be a good idea, I didn't want to suggest that the component handles the "close" part. But now that you mention it, I realize that we already demo some components this way so, I guess it's ok

gcornut (approved):

gcornut:
Test OK ✔️

FlorentDinet:
Test ok ✔️

PR #1115, opened 2024-03-28
chore(popover): allow to set the element to set a focus trap on

Fafnor:
# General summary
Add a focusZoneElement prop on the Popover component allowing to specify the element in which the focus trap should be applied.
Make focus trap works with element with shadow root

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   DatePicker: add an input to change the displayed year
 -   Message: add a `closeButtonProps` prop to add a close button in the message. Only available for `info` kind messages with a background.
+-   Popover: add a `focusTrapZoneElement` prop to specify the element in which the focus trap should be applied.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -16,10 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   DatePicker: add an input to change the displayed year
 -   Message: add a `closeButtonProps` prop to add a close button in the message. Only available for `info` kind messages with a background.
+-   Popover: add a `focusTrapZoneElement` prop to specify the element in which the focus trap should be applied.
 
 ### Changed
 
 -   Dialog: fullscreen display on smaller viewport
+-   FocusTrap: Make the focus trap work on element with shadow root
```
gcornut, line 24: Don't need to mention that in the changelog because the lumx libs do not expose the focus trap utility.
We only list changes that affect the the outside world

```suggestion
```

gcornut (approved):

gcornut:
Non regression tests 
- PopoverDialog focus trap ✅ 
- Dialog focus trap ✅ 
- Lightbox focus trap ✅

PR #1122, opened 2024-04-16
refactor(icons): optimize SVG and flatten output TS file

gcornut:
https://lumapps.atlassian.net/browse/DSW-171

# General summary

**Refactor our generate icons workflow:**
- Added optimization of our SVG files (in packages/lumx-icons/override/override-icons)
- Added optimization of all SVG paths (even from MDI)
- Changed output TS file
  - Flatten all icons into a single file re-defining MDI SVG as const (optimized)
  - Remove direct dependency to @mdi/js (now a devDependency)
- Added output icon-library.json file to use in our documentation site 

**Before**: dist/index.js => 2.1 M
**After**: dist/index.js => 1.6 M


hchevallier (approved):

gcornut:
- Fixed icon name convertion issue (conversion between "kebab-case" and "camel-case")
- Simplified icon library JSON format (grouped icon name, path and aliases into object in an array)
- Generate JS instead of TS (types still generate using the rollup build 👍)

PR #1124, opened 2024-04-17
chore(dependencies): fix eslint dev package version requirement

gcornut:
# General summary

Removing the eslint dev package from root package.json as it breaks the release workflow and it's not actually required


gcornut:
Local Lint OK ✅ 
CI Lint OK ✅

PR #1129, opened 2024-05-13
feat(flag): allow any react node as label and add truncate option

gcornut:
- **feat(flag): allow any react node in label**
- **feat(flag): add truncate text option on overflow**

# General summary

# Screenshots


packages/lumx-react/src/components/flag/Flag.tsx
```diff
@@ -28,17 +30,19 @@ const DEFAULT_PROPS: Partial<FlagProps> = {
  * @return React element.
  */
 export const Flag: Comp<FlagProps, HTMLDivElement> = forwardRef((props, ref) => {
-    const { label, icon, color, className, theme, ...forwardedProps } = props;
+    const { label, icon, color, className, theme, truncate, ...forwardedProps } = props;
```
hchevallier, line 33: Naive question, why don't you use a default value `false` here to avoid using `!!truncate` then?

hchevallier (approved):
Good for me, just a code styling question

packages/lumx-react/src/components/flag/Flag.tsx
gcornut: But I should have done that in the classes too 🤔

hchevallier: But it's typed as boolean, they can't unless they ignore the typing issues

PR #1105, opened 2024-03-19
fix(color-a11y): changed red and green, changed low and medium text color

FlorentDinet:
# General summary

To fix a contrast issue on red and green buttons:

1. red/D2, red/N and green/N color values are changed.
2. Colored Low and medium button texts are now using the D2 color instead of N color.

# Screenshots

image
image
image
image

StoryBook: https://e5f9ee1c--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=347)) **⚠️ Outdated commit**

PR #1131, opened 2024-05-16
fix(color): add fallback to default variant

armasson:
# General summary

Fix an issue introduce in this PR https://github.com/lumapps/design-system/pull/1105, we have issues with the change introduce here : https://github.com/lumapps/design-system/pull/1105/files#diff-e9c65473a041d47079ec7b0eb7892fa6f29a0be893a1fa17138777d5a16f1ac5R87 that's trying to apply a `D2` variant, to colours that does not possess this variant (`light` / `dark`, ...). 

To fix this issue there is multiple solution, we could revert / condition the fixe that has been made, to only apply the D2 only to the "Green" / "Red" colour, like it was meant in the first change, but that does not seem to be a strong fix, to add specific if condition based on the colour.

I've opted to a more global solution, and patching the `lumx-color-variant` function, to add a fallback to the default "N" variant, that should be in theory present in all colours. 

If we want to add an extra layer of safety, in case there is colours without the N variant, we could add another fallback to the dark colour, but it seems a bit much for a use case that will most likely never happened.

# Screenshots

StoryBook: https://1dfbf592--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=369))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+-   `@lumx/core`: lumx-color-variant, add fallback to default variant
+
 ## [3.7.0][] - 2024-04-29
 
 ### Added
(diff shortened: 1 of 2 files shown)


gcornut (approved):

gcornut:
Test OK - Stories visual diff before vs after look good ✅

PR #1136, opened 2024-06-07
fix(chip): forward key down event on clickable chip

gcornut:
DSW-191

Fix forwarding key down event on clickable chip


hchevallier (approved):
LGTM

PR #1142, opened 2024-06-20
fix(generic-block): fix shrinking of the content section

gcornut:
# General summary

Fix shrinking the generic block content instead of overflowing


juanigalan91 (approved):

gcornut:
| Before | After |
| --- | --- |
| image | image

PR #1151, opened 2024-08-06
fix(lightbox): fix a11y color contrast on the close button

gcornut:
# General summary

Fix a11y color contrast on the Lightbox close button

DSW-198

Before | After
 --- | --- 
image | image


gcornut:
Test OK (local SB) ✅

PR #1155, opened 2024-08-12
fix(heading): fix default typography when ̀`as` is set

hchevallier:
# General summary

When the `as` prop was set on `Heading` component, the default `typography` was not set correctly. It was only checking the calulated heading level, coming from an hypothetic HeadingLevelContext. If the context was not provided, it was using `h1` typography even for other levels.

# Screenshots
|Before   |After   |
|---|---|
|  image |image |


gcornut (approved):

gcornut:
Test OK (local SB) ✅

PR #1157, opened 2024-08-13
chore(release): add step to deprecate older prerelease version

gcornut:
# General summary

Add a step in release workflow to deprecate the older version of the prerelease after the new version has been uploaded

Testing:
1. Should not deprecate any version when using a prerelease tag that does not have an older version (ex: https://github.com/lumapps/design-system/actions/runs/10366047422/job/28694448631)
2. Should deprecate the old version when creating a new version for a prerelease tag (ex: https://github.com/lumapps/design-system/actions/runs/10366097796/job/28694607610)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,7 +89,8 @@ jobs:
                         console.log(`NPM dist tag: ${distTag}`);
 
                         // Get base version (version from which to increment)
-                        let baseVersion = await getVersion(distTag)
+                        const distTagBaseVersion = await getVersion(distTag);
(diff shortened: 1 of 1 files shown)


gcornut:
Test OK


1. Should not deprecate any version when using a prerelease tag that does not have an older version (ex: https://github.com/lumapps/design-system/actions/runs/10366047422/job/28694448631) ✅ 
2. Should deprecate the old version when creating a new version for a prerelease tag (ex: https://github.com/lumapps/design-system/actions/runs/10366097796/job/28694607610) ✅

hchevallier (approved):

PR #1159, opened 2024-08-13
chore(workflow): deploy chromatic on push master instead of tag

gcornut:
# General summary

Deploy chromatic when pushing the release commit on master rather than when pushing the release tag.

This should help chromatic know which reference to take (master) when comparing snapshots


gcornut:
Test OK
- job cancels if the commit is not a release commit ✅ 
- job continues otherwise ✅

hchevallier (approved):

PR #1160, opened 2024-08-13
chore(workflow): skip step rather than cancel job is deploy chromatic

gcornut:
# General summary

Skip steps rather than canceling the entire workflow to avoid cancellation notifications.

Use git command instead of GH API as it is more reliable (less server down time)


gcornut:
Test OK
- job skips steps if the commit is not a release commit ✅ 
- job continues otherwise ✅

hchevallier (approved):

PR #1161, opened 2024-08-13
chore(slideshow): fix small error in pagination items width

gcornut:
# General summary

Fixed a small oversight from https://github.com/lumapps/design-system/pull/1149 where the focus outline would still be cropped when selecting the right-most pagination item.

(not updating the CHANGELOG because there is already an entry talking about the fix from #1149)


gcornut:
Test OK
With 5 slides, the pagination take the most space but the sufficient space is reserved for the focus outline
Above or below 5 slides, focus outlines are not cropped either

hchevallier (approved):

PR #1163, opened 2024-08-14
fix(image-block): restore old caption classes and add props forwarding

gcornut:
# General summary

Restore old class names of iamge block caption elements.

Add title & description props forwarding.


hchevallier (approved):

gcornut:
Test OK ✅

PR #1180, opened 2024-09-06
fix(tooltip): fix re-render errors when removing the label

gcornut:
Remove render short-circuit when label is falsy to avoid react hook re-render errors


hchevallier (approved):

gcornut:
Test OK ✅

PR #1181, opened 2024-09-06
fix(image-lightbox): minor fixes

gcornut:
- fix(image-lightbox): update props each time the lightbox opens
- fix(image-lightbox): fix reset zoom scale when switching to the first item

StoryBook: https://ede04ee6--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=393))


hchevallier (approved):

gcornut:
Test OK
- Zoom is correctly reset when switching to the first slide ✅ 
- "Update props" fix is not directly testable here but at least it doesn't produce any regression ✅

PR #1183, opened 2024-09-16
fix(image-lightbox): fix zoom on mouse wheel cursor position

gcornut:
# General summary

pageY & pageX we used for the cursor position for the zoom center actually depended on the page scroll position but we wanted the cursor position in the window (since the image lightbox is always fullscreen)


hchevallier (approved):

gcornut:
Test OK (tested in an alpha release directly in the product) ✅

PR #1191, opened 2024-09-30
fix(image-lightbox): fix closing animation cut short because of unstable image reference

gcornut:
# General summary

Fix closing animation cut short because of unstable image reference


hchevallier (approved):

gcornut:
Test OK image lightbox close animation works properly now

PR #1193, opened 2024-10-02
chore(tooltip): update tooltip position on open

gcornut:
# General summary

We need to update the tooltip position on open for closeMode=hide

I'm not adding an entry in the CHANGELOG because it's technically a fix but for something that is not released yet


hchevallier (approved):

gcornut:
Test OK ✅

PR #1194, opened 2024-10-03
feat(tooltip): increment z-index above popover default z-index

gcornut:
# General summary

With the new mode closeMode=hide, the tooltip can appear below popovers because it can mount before the parent popover in the DOM

Updating the tooltip z-index be just above the default popover z-index


hchevallier (approved):

gcornut:
Test OK (on local SB)

PR #1198, opened 2024-10-09
fix(image-lightbox): fix closing transition triggering multiple times

gcornut:
# General summary

Fix bug where the image lightbox closing transition would fire multiple time when closing via "click away"


gcornut:
Test OK
Open/Close transition cancel correctly
Works well on Chrome & Safari (without any error in the console)

PR #1204, opened 2024-10-24
fix(expansion-panel): fix children remaining in the DOM when closed

Mcourt:
DSW-290

# General summary

When the expansion panel was closed, the children remained in the DOM and unhidden.
This could cause issues if content of expansion panel were interactive (such as inputs / buttons etc).

Fixed by removing the element from the DOM once the close animation finished.

Also checks the "reduced-motion" preference to immediately remove the children.

# Screenshots

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


gcornut (approved):

gcornut:
Test OK (tested in SB and in the demo site)
Expansion panel content unmounts correctly when closed ✅

PR #1207, opened 2024-11-04
fix(image-lightbox): windows 10 reporting incorrect pointer device

gcornut:
DSW-291


hchevallier (approved):

gcornut:
Can't really find exactly how to reproduce but the fix should target the original error 
`TypeError: Cannot read properties of undefined (reading &#x27;pointerId&#x27;)`

PR #1210, opened 2024-11-06
fix(expansion-panel): fix children remaining in DOM when isOpen changes

Mcourt:
DSW-294

# General summary

Due to a recent change, the `ExpansionPanel` was only removing the children from the DOM when clicking on the header/toggle button.
However, we have some cases where the panel closes without the header receiving a click (ex: when another panel is opened).

In these cases, the children remained in the DOM.

Fixed by using the `useVisibilityTransition` hook.

A `Close all` button has been added on a story to properly test it.

# Screenshots

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   `Chip`: trigger onClick when `Enter` key is pressed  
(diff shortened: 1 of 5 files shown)


gcornut (approved):

gcornut:
rebased and fix conflict on CHANGELOG

gcornut:
Test OK (tested locally on SB)
Non regression test OK: expansion panel is mounted on open and unmounted on close after the transition

PR #1219, opened 2024-11-29
fet(button): rework selected state style for better a11y

gcornut:


hchevallier (approved):
LGTM

gcornut:
Tested directly in the product with an alpha release 
Test OK ✅

PR #1220, opened 2024-11-29
feat(chip): rework selected state style for better a11y

gcornut:


gcornut:
Move to https://github.com/lumapps/design-system/pull/1219

PR #1221, opened 2024-11-29
feat(navigation): rework selected state style for better a11y

gcornut:


gcornut:
Moved to https://github.com/lumapps/design-system/pull/1219

PR #1223, opened 2024-12-04
fix(navigation-item): fix border left & right impacting the width

gcornut:
Rework the navigation item styles to make sure the border left & right does not affect the width

- Refactor navigation item styles to use "private" css variables to indirectly apply the public variables (design tokens)
- Calculate the left padding by removing the current border left width
- Calculate the right padding by removing the current border right width
- Make sure the border width default to `0px` and not `0` (or else it breaks the `calc()`)

StoryBook: https://3ec7fb6e--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=414))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
--   `Navigation`: fix focus outline in dark theme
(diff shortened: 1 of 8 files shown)


hchevallier (approved):

gcornut:
Test OK 
- Test chromatic ✅ 
- Test in storybook editing CSS border variables ✅

PR #1225, opened 2024-12-04
feat(button): add support for `isSelected` in `low` emphasis

gcornut:
- Reworked button styles so that the selected state can apply on top of both `medium` and `low` emphasis
- Deprecate selected state padding variables as we prefer to keep the base emphasis style for sizing

Selected state in low and medium emphasis is testable here: https://4436b6bb--5fbfb1d508c0520021560f10.chromatic.com/?path=/story/lumx-components-button-button--state-variations

StoryBook: https://1e0d3d2e--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=416)) **⚠️ Outdated commit**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+-   `Button`: add support of `isSelected` in `low` emphasis (in addition to `medium`).
+
 ### Changed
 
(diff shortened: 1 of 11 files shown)


gcornut:
Test OK 
(tested in product directly)
Button medium selected work like before
Button low selected work like before (but future proofed for the change on the border of the selected state)
(tested on product components: v1 follow button, wrex toolbar elements, v2 follow button)

hchevallier (approved):

PR #1235, opened 2025-01-09
docs(button): add danger and warning use cases

gcornut:
Add documentation for colored button use cases: `warning` and `danger`

Capture d’écran 2025-01-09 à 16 48 12


gcornut:
We'll be waiting for the semantic color palette rework before documenting them

PR #1237, opened 2025-01-10
fix(tabs): revert problematic flex-grow style on tablist

gcornut:


gcornut:
Test OK (manual test in storybook) ✅

PR #1262, opened 2025-02-10
fix(popover): auto update on anchor and popover resize

gcornut:
# General summary

Update popover placement on both anchor and popoverr resize


hchevallier (approved):

gcornut:
Test OK ✅ 
- Tested in SB
- Tested in alpha version in the product

PR #1268, opened 2025-02-25
fix(tooltip): fix overflow issue on closeMode hide

gcornut:
# General summary

Remove Tooltip placement style when hidden (closeMode=hide) to fix overflow issue


gcornut:
Tested in an alpha release in the product ✅

hchevallier (approved):

PR #1269, opened 2025-02-26
fix(mosaic): fix parent overflow issue caused by negative inset

gcornut:
# General summary

remove negative inset that could break overflow of parent element

StoryBook: https://e197ae44--5fbfb1d508c0520021560f10.chromatic.com/ ([Chromatic build](https://www.chromatic.com/build?appId=5fbfb1d508c0520021560f10&number=440))


gcornut:
Tested in an alpha release in the product ✅

gcornut:
Test KO => gap between images should be 2px

gcornut:
Test OK

PR #1270, opened 2025-03-03
fix(popover): limit popover placement update on element resize

gcornut:
# General summary

limit popover placement update on element resize


gcornut:
Tested in alpha release in the product (fixes trembling nested popover when they have a scroll)
(also tested with 20x CPU slowdown)

hchevallier (approved):
LGTM

PR #1275, opened 2025-03-14
feat: set default heading level context to all dialogs to h2

gcornut:
DSW-393

- **feat(dialog): set default heading level context to `h2`**
- **feat(lightbox): set default heading level context to `h2`**
- **feat(popover-dialogg): set default heading level context to `h2`**


hchevallier (approved):
LGTM

PR #1285, opened 2025-03-25
Filter and sort pattern documentation update

FlorentDinet:
# General summary

Update the documentation to reflect latest changes:
- use of a search field
- removed "exposed choices" use-cases as it is not used
- new responsive behavior
- precisions on facets and filter button usage

# Screenshots

Capture d’écran 2025-03-25 à 15 01 31


packages/site-demo/content/product/patterns/filter-and-sort/assets/filter-and-sort - anatomy.png
gcornut, line 1: Shouldn't it be cropped better ?

packages/site-demo/content/product/patterns/filter-and-sort/assets/filter-and-sort - sorting.png
gcornut, line 1: Same here, can we crop this image more on the right ?

packages/site-demo/content/product/patterns/filter-and-sort/assets/filter-and-sort - exposed facets.png
gcornut, line 1: The grey gaps a bit weird. Also maybe it could be better to split each demo into seperate files 
The gaps between the demos could be introduced via a flexbox in the docs

Something like:
```
<FlexBox orientation="vertical" gap="huge">
    <img src=demo1 />
    <img src=demo2 />

packages/site-demo/content/product/patterns/filter-and-sort/assets/filter-and-sort - responsive.png
gcornut, line 1: Same here I think the demos should be split and cropped individually

gcornut (changes requested):

packages/site-demo/content/product/patterns/filter-and-sort/assets/filter-and-sort - responsive.png
FlorentDinet: Ideally images should be replaced by demos, and this documentation moved to LumX private documentation as it uses privates components (search field, filters, sorter...)

gcornut (approved):

PR #1286, opened 2025-03-25
fix(link): fix parent typo not correctly inherited on button-like links

gcornut:
# General summary

fix parent typography not correctly inherited on button-like links


gcornut:
Moved to https://github.com/lumapps/design-system/pull/1287

PR #1291, opened 2025-04-02
fix(tabs): fix state update loop on React 18

gcornut:
# General summary

Fix tabs state update loop on React 18


Mcourt (approved):

gcornut:
Tested on product in alpha (react 18) ✅ Tested in SB (react 17) ✅

PR #1295, opened 2025-04-04
fix(link): fix focus outline warping around nested icons

gcornut:
# General summary

Fix weird warping of the link focus outline around nested icons

| Before | After |
| --- | --- |
| before | after |


hchevallier (approved):

gcornut:
Test OK 👍

PR #1299, opened 2025-04-23
feat(chip): add `xxs` size

gcornut:
# General summary

Add `xxs` size to avatar  
(ideal size to nest inside chips after/before)


hchevallier (approved):

gcornut:
Test OK

PR #1301, opened 2025-04-29
fix(icon-button): fix forward theme to icon

gcornut:
# General summary

Fix forward theme from IconButton to Icon


gcornut:
Test OK

PR #1302, opened 2025-05-07
fix(expansion-panel): allow content resize after opening transition

gcornut:
# General summary

fix allow content resize after opening transition


hchevallier (approved):
LGTM

gcornut:
Test OK

PR #1305, opened 2025-05-21
fix(thumbnail): handle 0 in focalPoint properly

DamienGarnes:
# General summary

Since `0 ` is a falsey value, fcal points equalling 0 in Thumnbail component were handled like null or undefined and skipped. But we actually need `0` to be handled like any other numeric value. This fix aims to avoid skipping 0, only `null` and `undefined` are now skipped.
 (other falsey values should be avoided thx to Typescript)
# Screenshots


gcornut (approved):

gcornut:
Test OK

PR #1306, opened 2025-05-21
feat(button): add disabled behavior with aria-disabled only

hchevallier:
# General summary

Make the button disableable with only the `aria-disabled` prop. 
If `aria-disabled` is set to `true`, the button will have all disabled behavior (prevent click, disabled styles, no hover or active styles) but displays the tooltip on hover correctly.

# Screenshots


gcornut:
Test OK ✅

PR #1310, opened 2025-06-16
fix(alert-dialog): improve a11y readings

armasson:
Remove "header" and "footer" HTML elements from the AlertDialog component to avoid confusion for the screen readers, and improve accessbility of the dialog, by providing the dialog context with "aria-describedby" and "aria-labelledby".

# General summary

This PR aim at improving the a11y of the "AlertDialog" component with screen readers. The current version of the AlertDialog is no accessible, and does not provide enough informations to the user, when the dialog open.

The `aria-describedby` and `aria-labelledby` attribute are correctly applied to the created `role="alertdialog"` element as well as the other required attribute `aria-modal="true"`. 

However, the current behavior do not make use of this attributes. Upon opening, the dialog set the focus on the footer (the first interactive element), and by doing so, skip the reading of the dialog attributes.

This is due to the semantic nature of the `footer` tag, that is supposed to contains enough semantic value so that the Dialog can be skipped. 

The fix then consist to remove `header` and `footer` tags, that does not seems to be required in order to keep a semantically valid component, and even more create some confusion for the assistive technologies. 

# Screenshots

### Before

https://github.com/user-attachments/assets/a8e60944-f32e-402b-b296-eb76267c54de

### After

https://github.com/user-attachments/assets/4adab4f7-dae2-4e36-a705-eeb076829e4c


gcornut:
Test OK ✅

PR #1653, opened 2026-05-06
fix(combobox): allow propagation of Enter key with no option active

gcornut:
# General summary

In a combobox input, key down event on `Enter` should propagate when no option is active in the listbox (key nav has not started, `aria-activedescendant` is empty)

StoryBook lumx-vue: https://91e6dfca9--697a023f84e832e23544fb3c.chromatic.com/

StoryBook lumx-react: https://91e6dfca9--5fbfb1d508c0520021560f10.chromatic.com/


Mcourt (approved):

gcornut:
Test OK ✅

PR #1655, opened 2026-05-07
feat(vue): create select button component

gcornut:
# General summary

Migrate SelectButton to vue

StoryBook lumx-react: https://349d7813b--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://349d7813b--697a023f84e832e23544fb3c.chromatic.com/

### DIFF
--- a/packages/lumx-vue/src/components/button/Button.tsx
+++ b/packages/lumx-vue/src/components/button/Button.tsx
@@ -1,5 +1,5 @@
 import isEmpty from 'lodash/isEmpty';
-import { computed, defineComponent, toRaw, useAttrs, useSlots } from 'vue';
+import { Comment, computed, defineComponent, toRaw, useAttrs, useSlots } from 'vue';
 
 import {
     Button as ButtonUI,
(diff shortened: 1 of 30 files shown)


packages/lumx-vue/src/components/button/IconButton.tsx
```diff
@@ -53,17 +53,24 @@ const IconButton = defineComponent(
...
 
-        // Ref to the underlying button DOM element, exposed so template refs resolve to the button.
-        const buttonRef = ref<HTMLElement>();
-        expose({ $el: buttonRef });
+        // Ref to the underlying button DOM element.
+        const buttonRef = ref<HTMLElement | null>(null);
+        const setButtonRef = (el: any) => {
+            buttonRef.value = (el?.$el ?? el) as HTMLElement | null;
+        };
+        expose({
```
durellinux, line 61: @andrzej-augustyniak I have no idea what this does... is it reactive written this way? And in general do we really care if this is reactive or not?

packages/lumx-vue/src/components/button/Button.tsx
```diff
@@ -59,15 +59,18 @@ const Button = defineComponent(
...
             // If single Text component, render directly
-            if (children.length === 1 && children[0].type === Text) {
-                return children[0];
+            if (visibleChildren.length === 1 && visibleChildren[0].type === Text) {
+                return visibleChildren[0];
             }
 
             // Otherwise wrap in span
-            return <span>{children}</span>;
+            return <span>{visibleChildren}</span>;
```
durellinux, line 73: What can be the children here? Can we wrap them always in a span?

gcornut: We do not really support anything else than text in button content for now

durellinux (approved):

PR #1661, opened 2026-05-11
chore(vue): add ClassValue type and strengthen class prop typing

juanigalan91:
## Summary

- Introduces `ClassValue` type in `@lumx/vue` (`VueToJSX.ts`) matching the recursive type accepted by Vue's `normalizeClass` and the `classnames` package
- Replaces `class?: string` with `class?: ClassValue` across all Vue components that expose the `class` prop, enabling arrays and objects in addition to strings
- Updates `useClassName` composable to accept `MaybeRefOrGetter<ClassValue>` instead of `string | undefined`
- Adds four class-management tests to `commonTestsSuiteVTL` covering: no-duplication on string class, array `ClassValue`, object `ClassValue`, and `class` + `className` merge — all components with `forwardClassName` inherit them automatically

## Affected files

- `packages/lumx-vue/src/utils/VueToJSX.ts` — new `ClassValue` export
- `packages/lumx-vue/src/composables/useClassName.ts` — updated parameter type
- `packages/lumx-vue/src/components/alert-dialog/AlertDialog.tsx`
- `packages/lumx-vue/src/components/dialog/Dialog.tsx`
- `packages/lumx-vue/src/components/image-block/ImageCaption.tsx`
- `packages/lumx-vue/src/components/radio-button/RadioGroup.tsx`
- `packages/lumx-vue/src/components/select-text-field/SelectTextField.tsx`
- `packages/lumx-vue/src/components/tooltip/Tooltip.tsx`
- `packages/lumx-vue/src/testing/commonTestsSuiteVTL.tsx` — 4 new class tests

## Testing

- All 1249 Vue unit tests pass
- Type-check clean (`yarn type-check:vue`)
- New tests verify string, array, and object class formats are applied without duplication, and that `class` + `className` merge correctly

StoryBook lumx-vue: https://3dd80a70c--697a023f84e832e23544fb3c.chromatic.com/

### DIFF
--- a/packages/lumx-vue/src/composables/useClassName.ts
+++ b/packages/lumx-vue/src/composables/useClassName.ts
@@ -2,6 +2,7 @@ import { type ComputedRef, type MaybeRefOrGetter } from 'vue';
 
 import { classNames } from '@lumx/core/js/utils';
 
+import { type ClassValue } from '../utils/VueToJSX';
 import { useAttrFallback } from './useAttrFallback';
 
 /**
@@ -17,6 +18,12 @@ import { useAttrFallback } from './useAttrFallback';
  * @param  classProp The Vue `class` prop value (or a getter/ref returning it).
  * @return Computed ref holding the merged class string (or undefined when empty).
  */
-export function useClassName(classProp: MaybeRefOrGetter<string | undefined>): ComputedRef<string | undefined> {
-    return useAttrFallback(classProp, 'className', classNames.join as any);
+export function useClassName(classProp: MaybeRefOrGetter<ClassValue>): ComputedRef<string | undefined> {
+    // `classNames.join` handles all ClassValue forms (string, array, object) natively.
+    // The outer cast is needed because useAttrFallback<ClassValue> can't infer string | undefined.
+    return useAttrFallback(
+        classProp,
+        'className',
+        (vue, fallback) => classNames.join(vue, fallback as string | undefined) || undefined,
+    ) as ComputedRef<string | undefined>;
 }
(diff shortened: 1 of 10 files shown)


packages/lumx-vue/src/composables/useClassName.ts
```diff
@@ -17,6 +18,6 @@ import { useAttrFallback } from './useAttrFallback';
  * @param  classProp The Vue `class` prop value (or a getter/ref returning it).
  * @return Computed ref holding the merged class string (or undefined when empty).
  */
-export function useClassName(classProp: MaybeRefOrGetter<string | undefined>): ComputedRef<string | undefined> {
-    return useAttrFallback(classProp, 'className', classNames.join as any);
+export function useClassName(classProp: MaybeRefOrGetter<ClassValue>): ComputedRef<string | undefined> {
+    return useAttrFallback(classProp, 'className', classNames.join as any) as ComputedRef<string | undefined>;
```
andrzej-augustyniak: Why all those `as` casts?

durellinux: Will this work? classProp can be an array or an object... will classNames.join work in that case?

Or does Vue when reaching this point already resolved the classProp to a single string/list of strings.

If it's the second, then it's worth it documenting it.

cc @andrzej-augustyniak

juanigalan91: Vue returns a resolved class as a string directly, I will improve the docs

juanigalan91: changing `string | undefined` to `ClassValue` made the return change types. I will tweak the code so we can avoid those casts

andrzej-augustyniak (approved):

durellinux (approved):

PR #1668, opened 2026-05-22
chore(lumx-vue): optional props for dialog

juanigalan91:
# General summary

- `Dialog` Vue component: `forceFooterDivider`, `forceHeaderDivider`, and `isLoading` props are now typed as optional via `Partial<Pick<...>>`, matching their intended optional nature in `BaseDialogProps`

# Screenshots

N/A — no visual change

StoryBook lumx-vue: https://0b4dc97e2--697a023f84e832e23544fb3c.chromatic.com/

### DIFF
--- a/packages/lumx-vue/src/components/dialog/Dialog.tsx
+++ b/packages/lumx-vue/src/components/dialog/Dialog.tsx
@@ -19,7 +19,7 @@ import { useTransitionVisibility } from '../../composables/useTransitionVisibili
 import { useDisableBodyScroll } from '../../composables/useDisableBodyScroll';
 import { keysOf, type ClassValue } from '../../utils/VueToJSX';
 
-export type DialogProps = Pick<BaseDialogProps, 'forceFooterDivider' | 'forceHeaderDivider' | 'isLoading'> &
+export type DialogProps = Partial<Pick<BaseDialogProps, 'forceFooterDivider' | 'forceHeaderDivider' | 'isLoading'>> &
     HasCloseMode & {
         /** Additional class name. */
         class?: ClassValue;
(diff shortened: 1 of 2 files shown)


packages/lumx-vue/src/components/dialog/Dialog.tsx
```diff
@@ -19,7 +19,7 @@ import { useTransitionVisibility } from '../../composables/useTransitionVisibili
 import { useDisableBodyScroll } from '../../composables/useDisableBodyScroll';
 import { keysOf, type ClassValue } from '../../utils/VueToJSX';
 
-export type DialogProps = Pick<BaseDialogProps, 'forceFooterDivider' | 'forceHeaderDivider' | 'isLoading'> &
+export type DialogProps = Partial<Pick<BaseDialogProps, 'forceFooterDivider' | 'forceHeaderDivider' | 'isLoading'>> &
```
gcornut, line 22: I don't think this changes anything. Those props are already optional in `BaseDialogProps`

And it's weird we don't have this bug in the vue demos from site-demo which are type checked with vue-tsc 🤔

gcornut: But if you can confirm this works (with an alpha version tested in the projects that has this bug) then it's good for me

juanigalan91: I'll check with an alpha sure, here's what happens on other projects, if you have other ideas I am all ears

Screenshot 2026-05-22 at 12 15 20

juanigalan91: Well, yeah, you were right, still buggy. Will keep searching i GUESS

PR #1671, opened 2026-05-25
chore(input-label): add as prop

juanigalan91:
# General summary

- Adds an `as` prop to `InputLabel` (core + Vue) to allow rendering as a `<legend>` instead of the default `<label>` — useful inside `<fieldset>` elements
- Core: destructures `as` and uses it as the tag, defaulting to `'label'`
- Vue: registers `as` in `keysOf` so it flows through as a prop
- Test: adds a shared test asserting the rendered element tag when `as="legend"` is passed; fixes the assertion (was incorrectly comparing DOM element to a string)

# Screenshots

StoryBook lumx-react: https://3c5886580--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://3c5886580--697a023f84e832e23544fb3c.chromatic.com/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   `@lumx/vue`:
     -   Create the `SelectButton` component
+       `InputLabel`: Added prop `as` to specify the HTML element to use for rendering
 -   `@lumx/react`:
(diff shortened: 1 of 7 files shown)


gcornut (approved):

juanigalan91:
- `InputLabel` correctly displayed with `legend` on React
- `InputLabel` correctly displayed with `legend` on Vue

PR #1674, opened 2026-05-27
feat(icon-button): also render label as hidden text inside the button

gcornut:
# General summary

Render the IconButton's label as a visually-hidden text node inside the button (in addition to keeping the existing `aria-label` for backward compatibility).

This will be useful when we want the button to also label another element. For example, we could have a dropdown menu attached to a button which is labeled by the button with `aria-labelledby={buttonId}`.

StoryBook lumx-react: https://922ba94f0--5fbfb1d508c0520021560f10.chromatic.com/
StoryBook lumx-vue: https://922ba94f0--697a023f84e832e23544fb3c.chromatic.com/


gcornut:
Might want to revisit that later. Closing for now

PR #1677, opened 2026-05-27
chore(select-button): minor improvements and fixes

gcornut:
### Fixes (`@lumx/core`, `@lumx/react`, `@lumx/vue`)

- **SelectButton: label fallback to `getOptionId`** — When `getOptionName` is not provided, the button label now falls back to `getOptionId` instead of rendering empty text. Uses the shared `getOptionDisplayName` utility.
- **SelectButton: Escape key no longer clears selection** — In button-mode (no text input), pressing Escape now only closes the dropdown without clearing the current selection. The `onChange` type is updated to never pass `undefined` (i.e. `(newValue: O) => void` instead of `(newValue?: O) => void`).
- **SelectButton: menu-down icon with `as=Button`** — Previously, passing `as={Button}` explicitly would skip the default chevron icon. Now the menu-down icon renders whenever `as` is `Button` or omitted.
- **ComboboxButton: fix ref change re-initialization** — Replaced `useRef` with `useState` for the button DOM element so that `setupComboboxButton` is re-invoked when the ref changes (e.g. on re-render with a new trigger element).

### Type improvements (`@lumx/react`)

- **`SingleSelectButtonProps` / `MultipleSelectButtonProps`** — Wrapped with `NamedProps` to strip the `GenericProps` index signature from the public aliases, giving consumers cleaner type-checking.
- **`TriggerProps`** — Uses `React.ComponentProps<E>` instead of the deprecated `React.ComponentPropsWithoutRef<E>`.

### Renames & API housekeeping (`@lumx/core`, `@lumx/react`, `@lumx/vue`)

- **`SelectTextFieldStatus` → `SelectListStatus`** — Renamed to reflect that the type is shared between `SelectButton` and `SelectTextField`. Backwards-compatible aliases (`SelectButtonStatus`, `SelectTextFieldStatus`) are re-exported from both React and Vue packages.
- **`getSectionId` accepts `undefined`** — The `getSectionId` selector and `renderSectionTitle` callback now accept `string | undefined`, allowing options without a section.

## Testing

- Unit test added: button label falls back to `getOptionId` when `getOptionName` is not provided
- Type-level tests updated: `onChange` no longer accepts `undefined`, `NamedProps` strips index signature, `Button` props (`isDisabled`, `emphasis`, `size`, `leftIcon`) are accessible on default trigger aliases
- Existing `commonTestsSuiteRTL` / `commonTestsSuiteTL` suites cover standard behavior

StoryBook lumx-react: https://d0edfb879--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://d0edfb879--697a023f84e832e23544fb3c.chromatic.com/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   `@lumx/react`:
     -   `SelectButton`: fix `SingleSelectButtonProps` and `MultipleSelectButtonProps` generic type
+    -   `SelectButton`: render menu-down icon with `as=Button` (same as when `as` is empty)
 -   `@lumx/react`, `@lumx/vue`:
     -   `SelectButton`, `SelectTextField`: allow undefined section id in `getSectionId`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+-   `@lumx/react`:
+    -   `SelectButton`: fix `SingleSelectButtonProps` and `MultipleSelectButtonProps` generic type
 -   `@lumx/react`, `@lumx/vue`:
     -   `SelectButton`, `SelectTextField`: allow undefined section id in `getSectionId`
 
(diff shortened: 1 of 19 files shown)


juanigalan91 (approved):

gcornut:
Test OK ✅

PR #1680, opened 2026-06-04
chore(combobox): rework render when closed

gcornut:
# Summary

Reworks the Combobox to only render its option list when the combobox is open, improving performance and DOM cleanliness. Also refactors focus navigation utilities and adds deferred navigation support for keyboard actions that open the combobox.

### Combobox conditional render
- **Combobox core**: Modify `setupCombobox` to conditionally render the options list based on open state. The list is only mounted when the combobox is expanded — both React and Vue wrappers.
- **ComboboxHandle**: Add `flushPendingNavigation()` to replay navigation intents that were deferred because options weren't in the DOM yet.
- **React/Vue wrappers**: Call `flushPendingNavigation()` via `useLayoutEffect`/`watch` after options commit on open.
- **Simplified arrow key logic**: ArrowDown/ArrowUp no longer guard on `hasNavigableItems` — they always open first, then navigate.
- **Removed `notify('optionsChange', ...)` on open** (was needed for empty list detection, no longer necessary with conditional render).

### Popover
- Set `hidden` attribute on the popover root when `closeMode="hide"` so it is excluded from the accessibility tree.

### Focus navigation utilities
- **Extract grid/list selectors**: Pure side-effect-free selection layer into `createGridSelectors`/`createListSelectors` files. Read-only accessors (`getFirst`, `getLast`, `getMatching`, `findNearestEnabled`) live on `nav.selectors`.
- **`createPendingNavigation`**: New utility for deferred navigation — stores an intent and retries on flush until it succeeds (or is cleared/aborted).
- **`goTo` resolver**: Replace `goToFirst`/`goToLast`/`goToItemMatching` with a unified `nav.goTo((s) => s.getFirst())` pattern that defers when the target is not yet in the DOM.
- **`FocusNavigationSelectors`** / **`ListFocusNavigationSelectors`** types: Public read interfaces exposed via `nav.selectors`.
- **`FocusNavigationController`**: Generic over `Selectors`, removing direct `activeItem`/`hasActiveItem`/`hasNavigableItems` from the controller in favor of `selectors`.
- **`setupRovingTabIndex`**: Updated to use `nav.selectors.*` and `nav.goTo((s) => s.getFirst())`.

### Typeahead
- **`createTypeahead`**: Add `rematch()` method that re-runs the match against the current search buffer without mutating it — used by button mode when navigation is deferred.
- **`getOptionLabel`**: New utility — typeahead matches against the visible label (`textContent`), not `data-value`, so the user types what they see.
- **`setupComboboxButton`**: Use `getOptionLabel` for typeahead walking, use `nav.goTo((n) => typeahead.rematch(n.activeItem))` for deferred matching.

### New utilities
- `lastDescendant`: Deepest last-child descendant of an element.
- `first`: Return first item of an iterable.

### Tests
- **Combobox core tests**: Open combobox before querying options (since options are no longer in DOM when closed). Migrate from raw `querySelector` to Testing Library queries (`screen.getByRole`, `getByClassName`, etc.).
- **Combobox React/Vue option tests**: Add `Combobox.Input` to test setups, open combobox before querying options, verify option is a native `<button>`.
- **SelectButton tests**: Open combobox before querying options.
- **Test stories**: Remove browser-only tests for auto-filter, `filter="off"`, and select-updates-input (covered by jsdom tests). Add new browser-only tests for button typeahead (from closed/while open), Home/End/ArrowDown from closed.
- **Claude.md**: Document testing conventions (prefer Testing Library queries, `find*` over `waitFor`).

### Type improvements and code cleanup
- Various type improvements across focus navigation and combobox.
- Remove `goToSelectedOrFirst`/`goToSelectedOrLast` utilities.
- Remove `goToItemMatching` from controller interface.

## Testing

- All core tests passing
- All React Combobox tests passing (unit + storybook)
- All Vue Combobox tests passing (unit + storybook)

StoryBook lumx-vue: [Deploying...](https://github.com/lumapps/design-system/actions/runs/27132658615?pr=1680)

StoryBook lumx-react: https://00ae97083--5fbfb1d508c0520021560f10.chromatic.com/

### DIFF
--- a/packages/lumx-core/src/js/components/Combobox/setupCombobox.ts
+++ b/packages/lumx-core/src/js/components/Combobox/setupCombobox.ts
@@ -1,12 +1,5 @@
 import { type FocusNavigationController } from '../../utils/focusNavigation';
-import {
-    getOptionValue,
-    goToSelectedOrFirst,
-    goToSelectedOrLast,
-    isActionCell,
-    isOptionDisabled,
-    notifySection,
-} from './utils';
+import { getOptionValue, isActionCell, isOptionDisabled, isSelected, notifySection } from './utils';
 import { setupListbox } from './setupListbox';
 import type {
     ComboboxCallbacks,
@@ -191,10 +184,10 @@ export function setupCombobox(
 
             switch (event.key) {
                 case 'Enter':
-                    if (handle.isOpen && nav?.hasActiveItem && nav.activeItem) {
+                    if (handle.isOpen && nav?.selectors.activeItem) {
                         // Capture activeItem before click — the click handler may close
                         // the popover and clear the focus navigation state.
-                        const { activeItem } = nav;
+                        const { activeItem } = nav.selectors;
                         // "Click" on active option
                         if (!isOptionDisabled(activeItem)) {
                             activeItem.click();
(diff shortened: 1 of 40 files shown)


packages/lumx-core/src/js/components/Combobox/setupCombobox.ts
```diff
@@ -214,40 +207,36 @@ export function setupCombobox(
...
+                            if (nav.type === 'grid') {
+                                nav.goDown();
+                            } else if (!nav.goToOffset(1) && wrapNavigation) {
+                                nav.goTo((s) => s.getFirst());
                             }
                         } else {
-                            // Open the listbox and focus selected option, fall back to first.
-                            handle.setIsOpen(true);
-                            if (!altKey) goToSelectedOrFirst(nav);
+                            nav.goTo((s) => s.getMatching(isSelected) ?? s.getFirst());
```
juanigalan91, line 221: Maybe some comments here to explain what is going on would be acceptable

packages/lumx-core/src/js/utils/focusNavigation/createGridSelectors.ts
juanigalan91, line 1: should we add some tests here?

gcornut: This one function does not make sense outside createGridFocusNavigation (which itself is tested)
Sooo I either I don't test or I move it inside createGridFocusNavigation 🤔

juanigalan91 (approved):

gcornut:
Testing in alpha... 🚧

gcornut:
Test OK ✅

PR #1682, opened 2026-06-09
chore(publish): remove deprecated publishConfig.directory

gcornut:
# General summary

npm v11+ warns about unknown publishConfig keys and will error in the next major version. The directory config was redundant since the CI publish workflow already cds into /dist before running npm publish. Build configs updated to hardcode 'dist' instead.

To test that we'll publish an alpha version from this branch

StoryBook lumx-react: https://fc8eb6eb7--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://fc8eb6eb7--697a023f84e832e23544fb3c.chromatic.com/

### DIFF
--- a/packages/lumx-core/package.json
+++ b/packages/lumx-core/package.json
@@ -53,9 +53,6 @@
         "./css/*": "./css/*",
         "./scss/*": "./scss/*"
     },
-    "publishConfig": {
-        "directory": "dist"
-    },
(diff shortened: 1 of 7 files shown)


gcornut:
Moved to https://github.com/lumapps/design-system/pull/1683

PR #1694, opened 2026-06-16
feat(time-picker-field): add minDateTime and maxDateTime bound props [CF-1546]

gcornut:
# General summary

- Add `minDateTime` and `maxDateTime` bound props to TimePickerField
- New `snapTimeToBounds` utility in core to clamp time values to configurable bounds
- Wired through React and Vue wrappers

# Screenshots

N/A — API-only change, no visual impact

StoryBook lumx-vue: https://ac0516357--697a023f84e832e23544fb3c.chromatic.com/

StoryBook lumx-react: https://ac0516357--5fbfb1d508c0520021560f10.chromatic.com/

### DIFF
--- a/packages/lumx-core/src/js/components/TimePickerField/Tests.tsx
+++ b/packages/lumx-core/src/js/components/TimePickerField/Tests.tsx
@@ -62,6 +62,11 @@ export default function timePickerFieldTests({ components, renderWithState }: Ti
             expect(inputValue).toMatch(/^2:30/);
             expect(inputValue.toUpperCase()).toContain('PM');
         });
+
+        it('renders the input as disabled when isDisabled is true', () => {
+            renderWithState(defaultTemplate, { isDisabled: true });
+            expect(screen.getByRole('combobox')).toBeDisabled();
+        });
     });
 
     describe('option list', () => {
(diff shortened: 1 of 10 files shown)


packages/lumx-core/src/js/components/TimePickerField/index.tsx
```diff
@@ -51,19 +51,38 @@ export interface TimePickerFieldWrapperProps {
...
+    /**
+     * Lower bound (date part respected). Same behaviour as `minTime` but
+     * compares the full date-time formed by combining each option's time with
+     * the current value's date. Skipped when `value` is undefined (no date
+     * context).
+     *
+     * When set and the current value falls below this bound, the component
+     * auto-selects the bound value on mount and on bound/value changes.
+     */
+    minDateTime?: Date;
```
durellinux, line 74: I am confused... this is a TimePicker only right? What's the use case to consider also the date?

Is teh use case to have a date picker, followed by a time picker where `minDateTime` and `maxDateTime` are set depending on same policy and `value` is updated with the value of the date picker? In this way we can check that the full date time is within specific range?

If so I would at least rework the description here because it is not clear.
These two fields do not have the same behavior as minTime and maxTime, as those predicates on the available options on the picker, these two do not, or at least not in the same way.

i.e. if my interpretation is right you can have:
- minDateTime: now - 1 day at current time
- maxDateTime: now + 1 day at current time
- minTime: 00:00
- maxTime: 23:59

Tbh this one feels like a separate component that has both date and time picker into it.
When you change the date, minTime and maxTime are updated in the time picker according to the boundaries set on the date time picker.

packages/lumx-core/src/js/components/TimePickerField/Tests.tsx
```diff
@@ -161,6 +205,49 @@ export default function timePickerFieldTests({ components, renderWithState }: Ti
...
+            renderWithState(defaultTemplate, { value, maxDateTime, onChange });
+            const input = screen.getByRole('combobox');
+            await userEvent.click(input);
+            await userEvent.clear(input);
+            await userEvent.type(input, '18');
+            await userEvent.tab();
+            expect(onChange).toHaveBeenLastCalledWith(expectTimeOfDay(14, 0), undefined, undefined);
+        });
+
+        it('skips `minDateTime` snap when value is cleared (no date context)', async () => {
```
durellinux, line 236: I am not getting what this test is doing.
The `clear` part is in all hte other test too, why is this one different?

gcornut:
Will split this 
- a new boundsMode to control when we clamp the value to minTime/maxTime https://github.com/lumapps/design-system/pull/1701 
- a util to link date picker bounds to time picker bounds

PR #1696, opened 2026-06-18
feat(popover): add a complete sizing API [CF-1597]

gcornut:
## Summary

Add a complete sizing API to `Popover` with six new props (`width`, `minWidth`, `maxWidth`, `height`, `minHeight`, `maxHeight`) supporting pixel strings, viewport-relative units, and global t-shirt size tokens (`xxs`–`xxl`).

Also moves the Popover floating-ui utility functions from a monolithic `popoverStyle.ts` into `utils/` with separate modules, and extracts shared types (`Offset`, `Elevation`, `PopoverSizes`) into `types.ts`.

## Changes

- **`@lumx/core`**: New `PopoverSizes` interface; CSS utilities `resolveCssSize`, `cssMin`, `cssMax`; refactored `popoverStyle.ts` → `utils/` (6 files)
- **`@lumx/react`**: Wire sizing props through `Popover` → `usePopoverStyle` → core middleware
- **`@lumx/vue`**: Same wiring with reactive refs + emit schema
- **`ComboboxPopover`**: Accept and forward sizing props; default `maxHeight: "80vh"` replaces hardcoded SCSS `max-height: 50vh`
- **Stories**: New `Width` and `Height` story variants with t-shirt size combinations
- **Docs**: Reorganized "Matching width" section into "Sizing" with full sizing API documentation

StoryBook lumx-react: https://b6a1aacdf--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://b6a1aacdf--697a023f84e832e23544fb3c.chromatic.com/


juanigalan91 (commented):
==

packages/lumx-core/src/js/components/Popover/types.ts
```diff
@@ -0,0 +1,41 @@
...
+    away?: number;
+}
+
+/**
+ * Popover elevation index.
+ */
+export type Elevation = 1 | 2 | 3 | 4 | 5;
+
+/** Popover size value — pixel value, or "t-shirt" size token. */
+export type PopoverSize = PXSize | (typeof POPOVER_SIZES)[number];
```
juanigalan91, line 20: `PopoverSize` comes from `POPOVER_SIZES = [Size.m, Size.l, Size.xl, Size.xxl]` so TS will reject `xxs`, `xs` and `s` even though `resolveCssSize` handles them fine and the docs say they're supported. Intentional? If not, using `GlobalSize` directly would fix it.

packages/lumx-core/src/js/components/Combobox/constants.ts
```diff
@@ -0,0 +1,3 @@
+import { VHSize } from '../../utils/browser/css/types';
+
+export const DEFAULT_COMBOBOX_POPOVER_MAX_HEIGHT: VHSize = '80vh';
```
juanigalan91, line 3: The SCSS had `max-height: 50vh` but the new default is `80vh` — was that intentional or should it stay at `50vh`?

packages/lumx-react/src/components/popover/index.ts
```diff
@@ -1,2 +1,3 @@
 export { Popover, type PopoverProps } from './Popover';
-export { Placement, type Offset, type Elevation } from '@lumx/core/js/components/Popover/constants';
+export { Placement } from '@lumx/core/js/components/Popover/constants';
+export type { Offset, Elevation } from '@lumx/core/js/components/Popover/types';
```
juanigalan91: `PopoverSizes` (and friends) aren't exported from the package index — anyone wanting to type a wrapper around Popover would need to deep-import. Worth re-exporting from `lumx-react` and `lumx-vue` index files.

gcornut: Good idea I will export PopoverHeight and PopoverWidth (the others are internal)

packages/lumx-core/src/js/components/Combobox/constants.ts
gcornut: Yes it's intentional. Let me update the changelog to be clear on that

packages/lumx-core/src/js/components/Popover/types.ts
gcornut: Yes intentional Size.m is already pretty small for a popover. I don't think we need smaller
And resolveCssSize is a generic util that applies to all `GlobalSize`

juanigalan91 (approved):

gcornut:
Test OK ✅

PR #1709, opened 2026-06-24
chore(ci): optimize CI with affected package detection and checkout and setup tunning

gcornut:
### 1. Optimize check-affected checkout and setup
- Use `filter: blob:none`, `show-progress: false`, `persist-credentials: false` for faster checkout
- Use `focus: '@lumx/actions'` to install only the actions workspace (nx + @actions/core) instead of full install
- Use Nx `parseFiles` + `calculateFileChanges` instead of raw `git diff` for more accurate change detection
- New outputs: `affected-test` (projects with `test` target), `affected-test-storybook` (projects with `test:storybook` target)

### 2. Only run affected package tests
- `tests` job matrix now reads from `needs.check-affected.outputs.affected-test` instead of hardcoded `["'@lumx/core', '@lumx/react', '@lumx/vue']`
- Job is skipped when no affected packages have a `test` target

### 3. Only run affected package storybook tests
- `test-storybook` now reads `packages` input from `needs.check-affected.outputs.affected-test-storybook`
- Each matrix step conditionally runs based on `contains(fromJSON(inputs.packages), ...)`
- Visual diff report handles "not tested" packages gracefully

### 4. Regroup lint + type-check into single code-check job
- Merged separate `lint` and `type-check` jobs into one `code-check` job
- Runs `nx run-many -t type-check,lint --projects=$AFFECTED` targeting only affected packages
- Removes ~2min of duplicate checkout+setup overhead

### 5. Fix: remove focus install for code-check
`focus: '@lumx/lumx'` skipped installing `eslint-config-lumapps` (dev-packages), causing ESLint to fail with missing `"lumapps"` config. Code-check now does a full install — the main perf win is the targeted `nx run-many`.

### 6. Fix: bypass nx task graph in check-build-site
`nx run lumx-site-demo:build` hit a circular dependency (`@lumx/react` imports from site-demo via `.storybook/main.ts`). Changed to `yarn workspace lumx-site-demo build` directly.

## Test Results

All scenarios verified via temp commits pushed to this PR branch:

| Scenario | code-check | tests | test-storybook | check-build-site |
|---|---|---|---|---|
| **CI-only change** (no source) | ✅ Pass | ⏭️ Skip | ⏭️ Skip | ⏭️ Skip |
| **Core change** (`packages/lumx-core/src/`) | ✅ Pass | ✅ core+react+vue | ✅ react+vue | ⏭️ Skip |
| **Icons change** (`packages/lumx-icons/`) | ✅ Pass | ✅ core+react+vue | ✅ react+vue | ⏭️ Skip |
| **Site-demo change** (`packages/site-demo/src/`) | ✅ Pass | ⏭️ Skip | ⏭️ Skip | ✅ Pass |

affected-test correctly filters packages without `test` target (e.g., `@lumx/actions`, `@lumx/visual-diffs`, `lumx-site-demo`).<br>
required-checks correctly aggregations all jobs — failure in any required job cascades properly.

### DIFF
--- a/.github/actions/check-affected/index.mjs
+++ b/.github/actions/check-affected/index.mjs
@@ -15,12 +15,12 @@ async function main() {
     const { files: changedFiles } = parseFiles(nxArgs);
     const touchedFiles = changedFiles.length > 0 ? calculateFileChanges(changedFiles, nxArgs) : [];
 
-    const affectedGraph =
-        changedFiles.length > 0 ? await filterAffected(graph, touchedFiles) : graph;
+    const affectedGraph = changedFiles.length > 0 ? await filterAffected(graph, touchedFiles) : graph;
     const affectedNames = Object.keys(affectedGraph.nodes);
 
     // projects with at least one modified file
     const modified = [];
+    const affectedByTarget = { test: [] };
     for (const [name, node] of Object.entries(graph.nodes)) {
         const root = node.data.root;
         const prefix = root + '/';
@@ -28,12 +28,17 @@ async function main() {
         if (changedFiles.some((f) => f === root || f.startsWith(prefix))) {
             modified.push(name);
         }
+        if (node?.data?.targets?.['test']) affectedByTarget.test.push(name);
     }
 
     console.log('affected', affectedNames);
     core.setOutput('affected', JSON.stringify(affectedNames));
+
     console.log('modified', modified);
     core.setOutput('modified', JSON.stringify(modified));
+
+    console.log('affected-test', affectedByTarget.test);
+    core.setOutput('affected-test', JSON.stringify(affectedByTarget.test));
 }
 
 main().catch((err) => {
--- a/.github/actions/check-affected/index.mjs
+++ b/.github/actions/check-affected/index.mjs
@@ -16,19 +16,34 @@ async function main() {
     const touchedFiles = changedFiles.length > 0 ? calculateFileChanges(changedFiles, nxArgs) : [];
 
     const affectedGraph = changedFiles.length > 0 ? await filterAffected(graph, touchedFiles) : graph;
-    const affectedNames = Object.keys(affectedGraph.nodes);
 
-    // projects with at least one modified file
+    // Affected projects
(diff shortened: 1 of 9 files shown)


.github/actions/setup/action.yml
```diff
@@ -15,22 +19,28 @@ runs:
...
+                cache: yarn
 
         -   name: "Cache dependencies"
+            if: inputs.focus == ''
             id: cache
             uses: actions/cache@v5
             with:
-                path: |
-                  **/node_modules
-                  .yarn/cache
```
gcornut, line 25: .yarn/cache is automatically cached in actions/setup-node with cache: yarn

juanigalan91 (approved):

.github/actions/check-affected/index.mjs
```diff
@@ -31,10 +26,29 @@ async function main() {
         }
     }
 
+    // Affected projects that have a test target (to avoid running tests on non-testable packages).
+    const affectedWithTests = affectedNames.filter((name) => {
```
durellinux: This works, I would wonder if it won't be better to add a test target by default and skip failing on missing tests for the one that should not be tested.

.github/actions/check-affected/index.mjs
```diff
@@ -1,22 +1,17 @@
...
     const graph = await createProjectGraphAsync();
 
-    const touchedFiles = changedFiles.map((file) => ({
-        file,
-        getChanges: () => [new WholeFileChange()],
-    }));
+    const base = execSync('git merge-base origin/master HEAD').toString().trim();
+    const nxArgs = { base, head: 'HEAD' };
+    const { files: changedFiles } = parseFiles(nxArgs);
+    const touchedFiles = changedFiles.length > 0 ? calculateFileChanges(changedFiles, nxArgs) : [];
```
durellinux, line 14: Why are we not using the affected list directly?

Like running `nx affected -t test` below which takes care of running already only the projects with test target defined?

gcornut: Doing it that way I don't even spawn GH jobs if there is no affected package

PR #1722, opened 2026-07-15
feat(tabs): add nav-link mode rendering

gcornut:
# General summary

Adds a nav-link rendering mode to `Tab`/`TabList` (React + Vue), so a `TabList` can be used as a navigation menu (links to pages/routes) instead of an in-page tab panel switcher.

-   **Mode is driven by `TabProvider` presence, not by `as`**: inside a provider → classic WAI-ARIA `tab` mode (`role="tab"`, roving tabindex, `aria-selected`); outside a provider → nav-link mode (plain link, `role="navigation"` on the list, `aria-current`, no tab roles/roving tabindex).
-   `Tab` becomes polymorphic on `as` — this only selects the rendered element and applies in both modes. React: `as="a"` (with `href` required at compile time via `HasRequiredLinkHref`) or `as={RouterLink}`, ref type follows `as` (`forwardRefPolymorphic`). Vue: an `action` scoped slot supplies the element.
-   `TabList` resolves its container role from provider presence directly (`navigation` vs `tablist`) rather than scanning children; roving-tabindex hook is left inert in nav-link mode since nav links keep native per-link tab stops.
-   Active state in nav-link mode is the caller's responsibility (e.g. a router `NavLink` setting `aria-current="page"`); the demo `HashNavLink` shows this via URL hash.
-   Removed the earlier nav-link context/registration mechanism and the now-unused `tabId`/`useTabProviderContextState` plumbing in favor of the provider-presence check.
-   Docs: new "Navigation" demo (React + Vue) + accessibility notes on the Tabs page.

# Screenshots

StoryBook lumx-react: https://27eef50a8--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://27eef50a8--697a023f84e832e23544fb3c.chromatic.com/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+-   `@lumx/react`, `@lumx/vue`:
+    -   `Tab`: support rendering as a custom element — `as` prop in React, `action` slot in Vue — e.g. to render navigation links.
+
+### Changed
+
+-   `@lumx/react`, `@lumx/vue`:
+    -   `TabList`: render a `role="navigation"` landmark instead of `role="tablist"` when used outside a `TabProvider`.
+    -   `Tab`: render with WAI-ARIA `role="tab"` semantics only inside a `TabProvider`.
+
 ### Fixed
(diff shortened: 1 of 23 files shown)


Mcourt (approved):
Good for react but I can't really speak for vue.

PR #1726, opened 2026-07-17
feat(dialog): add DialogHeading for automatic accessible naming

gcornut:
CF-1720

# General summary

Adds a `DialogHeading` subcomponent (React + Vue) that gives `Dialog`, `Lightbox`, and `PopoverDialog` an accessible name automatically — no manual `aria-label(ledby)` wiring. Split into 6 commits, one per layer:

-   **`chore(utils)`: `IdsRegistry`** — new cross-component id/label registry (`createIdsRegistry` in `@lumx/core`, `IdsRegistryContext`/`IdsRegistryProvider`/`useRegisterId`/`useRegisteredId` in React and Vue) shared by every dialog-like component.
-   **`feat(dialog)`: `DialogHeading`** — new subcomponent (React + Vue); rendered inside a dialog-like container it registers its id in the registry, and the container applies `aria-labelledby` automatically. No heading → no attribute rendered. Explicit `aria-label`/`aria-labelledby` (incl. via `dialogProps`) always takes precedence.
-   **`feat(dialog)`: wire registry into `Dialog`** — core split into `DialogShell` (portal/theme/registry-provider) + `DialogContainer` (dialog UI), plumbed through the React/Vue wrappers. `AlertDialog`'s manual `aria-label` wiring is untouched.
-   **`feat(lightbox)`: `DialogHeading` support** — wrapper/Content split (React + Vue) so `LightboxContent` can read the registry below the provider; Vue gets co-located `constants.ts`/`types.ts`.
-   **`feat(popover-dialog)`: `DialogHeading` support** — same wrapper/Content split; `PopoverDialog.label` is now optional since a heading can provide the accessible name instead.
-   **`docs(site-demo)`**: React/Vue `Dialog`, `Lightbox`, and `PopoverDialog` demos migrated to `DialogHeading`.

Full unit + Storybook `play` coverage added for the registry and each component.

# Screenshots

StoryBook lumx-vue: https://bd84c3ded--697a023f84e832e23544fb3c.chromatic.com/

StoryBook lumx-react: https://bd84c3ded--5fbfb1d508c0520021560f10.chromatic.com/


packages/lumx-react/src/components/lightbox/LightboxContent.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+            // Focus the parent element on close.
+            if (!isOpen && parentElement && parentElement.current) {
+                parentElement.current.focus();
+            }
+        }
+    }, [isOpen, parentElement]);
+
+    // Close lightbox on escape key pressed.
+    useCallbackOnEscape(onClose);
```
juanigalan91: Should this be guarded by `preventAutoClose`? Right now Escape closes the Lightbox even when it's set. Vue's `LightboxContent` gates it (`isOpen && !preventAutoClose`), so the two seem to diverge on the same prop. Looks pre-existing, but might be worth aligning here — maybe `useCallbackOnEscape(onClose, isOpen && !preventAutoClose)`?

CHANGELOG.md
```diff
@@ -11,13 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
     -   `Tab`: support rendering as a custom element — `as` prop in React, `action` slot in Vue — e.g. to render navigation links.
+    -   `Dialog`: add `DialogHeading` component — names the enclosing `Dialog`, `Lightbox`, or `PopoverDialog` by linking it via `aria-labelledby` when rendered anywhere inside.
 
 ### Changed
 
 -   `@lumx/react`, `@lumx/vue`:
     -   `TabList`: render a `role="navigation"` landmark instead of `role="tablist"` when used outside a `TabProvider`.
     -   `Tab`: render with WAI-ARIA `role="tab"` semantics only inside a `TabProvider`.
     -   `InfiniteScroll`: allow `undefined` callback to disable it
+    -   `PopoverDialog`: `label` prop is now optional. rendering a `DialogHeading` inside is an alternative way to satisfy its accessible-name requirement.
```
juanigalan91, line 22: Should the Lightbox change land here too? It now emits only one of `aria-label`/`aria-labelledby` (label wins) instead of both, and `ariaLabel`/`ariaLabelledBy` look deprecated now — both seem worth a line, maybe a `Deprecated` section for the latter?

packages/lumx-vue/src/components/dialog/Dialog.tsx
```diff
@@ -153,40 +155,46 @@ const Dialog = defineComponent(
...
+                        dialogProps={props.dialogProps}
+                        footerChildContent={
+                            (slots.footer?.() ?? (footerVnode as any)?.children) as JSXElement | undefined
+                        }
+                        forceFooterDivider={props.forceFooterDivider}
+                        forceHeaderDivider={props.forceHeaderDivider}
+                        handleClose={handleClose}
+                        hasBottomIntersection={hasBottomIntersection.value}
+                        hasTopIntersection={hasTopIntersection.value}
+                        headerChildContent={
```
juanigalan91, line 199: React forwards the inline `<header>`/`<footer>`'s own props/class, but here only `headerChildContent`/`footerChildContent` get passed — so an inline `<header class="x">` would lose its class in Vue, right? `DialogContent` even declares `headerChildProps`/`footerChildProps`, they're just never filled. (Named-slot usage isn't affected.)

packages/lumx-react/src/components/popover-dialog/PopoverDialogContent.tsx
```diff
@@ -0,0 +1,61 @@
...
+            ref={ref}
+            className={classNames.join(className, CLASSNAME)}
+            role="dialog"
+            aria-modal="true"
+            /**
+             * If a label is set, set as aria-label.
+             * If it is undefined, the label can be set using the `aria-label` and `aria-labelledby` props,
+             * or by rendering a `DialogHeading` inside the dialog (linked via the internal ids registry).
+             */
+            {...resolveAccessibleNameProps(label, ariaLabelledByAttr || labelId)}
```
juanigalan91, line 53: Now that `label` is optional, a PopoverDialog with no `label`/`aria-*` and no `DialogHeading` ends up with no accessible name and nothing flags it. Worth a dev-only `console.warn` when nothing resolves? It's documented as intentional, so just a nice-to-have.

packages/lumx-react/src/utils/IdsRegistryContext/IdsRegistryContext.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+        expect(screen.getByTestId('reader')).toHaveAttribute('aria-labelledby', 'heading-1');
+
+        await userEvent.click(screen.getByRole('button', { name: 'hide' }));
+        expect(screen.getByTestId('reader')).not.toHaveAttribute('aria-labelledby');
+    });
+
+    it('should be a no-op outside any provider', () => {
+        render(<Reader />);
+        expect(screen.getByTestId('reader')).not.toHaveAttribute('aria-labelledby');
+    });
```
juanigalan91, line 85: Vue has a test for the registered id updating reactively — worth adding the React equivalent here (registrant's `id` changes → reader updates)?

packages/lumx-core/src/js/components/Dialog/DialogShell.tsx
```diff
@@ -0,0 +1,93 @@
...
+                className={classNames.join(
+                    className,
+                    block({
+                        'is-hidden': !isOpen,
+                        'is-loading': isLoading,
+                        'is-shown': isOpen || isVisible,
+                        [`size-${size}`]: Boolean(size),
+                    }),
+                )}
+                style={{ zIndex }}
```
juanigalan91, line 81: `style={{ zIndex }}` overwrites any forwarded `style` here (it's spread after `forwardedProps`) — intended, or should it merge? Internal-only, so low stakes.

juanigalan91 (commented):
Went through the changes — mostly questions, nothing blocking. Really like the ids-registry primitive and how cleanly the Dialog/Lightbox/PopoverDialog split came out. A few things inline.

packages/lumx-react/src/utils/IdsRegistryContext/IdsRegistryContext.ts
```diff
@@ -0,0 +1,16 @@
+import { createContext } from 'react';
+
+import { IdsRegistry } from '@lumx/core/js/utils/idsRegistry/createIdsRegistry';
```
juanigalan91: `import type` here? `IdsRegistry` is only used as a type, and the Vue side already does `import { type IdsRegistry }` — probably trips `consistent-type-imports`.

packages/lumx-react/src/components/dialog/DialogHeading.tsx
```diff
@@ -0,0 +1,41 @@
+import { Heading, HeadingProps } from '@lumx/react/components/heading';
+import { useId } from '@lumx/react/hooks/useId';
+import { forwardRef } from '@lumx/react/utils/react/forwardRef';
+
+import { Typography } from '@lumx/core/js/constants';
+import { DIALOG_LABEL_KEY, DIALOG_HEADING_COMPONENT_NAME } from '@lumx/core/js/components/Dialog/constants';
+
+import { useRegisterId } from '@lumx/react/utils/IdsRegistryContext';
+
+export interface DialogHeadingProps extends HeadingProps {}
```
juanigalan91: Empty interface — Vue uses `export type DialogHeadingProps = HeadingProps`; matching that here would dodge `no-empty-object-type`?

packages/lumx-react/src/components/lightbox/LightboxContent.tsx
gcornut: Good catch! It's a pre-existing bug but I'll fix it nonetheless.

CHANGELOG.md
gcornut: Added a changelog line for the accessible-name behavior change (label now wins instead of emitting both). On the deprecation: `ariaLabel`/`ariaLabelledBy` were already `@deprecated` on master, so I left them out of this PR's changelog.

packages/lumx-vue/src/components/dialog/Dialog.tsx
gcornut: Again a pre-existing bug but I'll fix also :)

packages/lumx-react/src/components/popover-dialog/PopoverDialogContent.tsx
gcornut: With the IdsRegistry we might log some false negative (where we haven't yet found the dialog heading). I'd rather just document it well and educate everyone on the importance of labelling dialogs

packages/lumx-core/src/js/components/Dialog/DialogShell.tsx
gcornut: Pre-existing behavior, I'll skip that for now

gcornut:
Test OK

juanigalan91 (approved):

PR #1744, opened 2026-08-17
fix(clickaway): ignore mouse presses landing on a scrollbar

cestoliv:
# General summary

A press on a scrollbar closes any open popover.

Click away detection asked a single question: does any registered element contain the event target. A
browser dispatches a real `mousedown` on the scrolling element when you press its scrollbar, and a
scrolling element is an ancestor of the popover it holds. An ancestor is never contained by its own
descendant, so the press counted as a click away and the popover closed. Wheel and trackpad
scrolling dispatch no `mousedown`, which is why they never showed the problem.

`isClickAway` now also asks where on the target the press landed. A press that lands in the target's
own scrollbar gutter is not a click away. The check lives in `@lumx/core`, so one change covers
`@lumx/react`, `@lumx/vue`, and every component built on them.

# See it

## Before the fix

-   React: https://f0c5197d6--5fbfb1d508c0520021560f10.chromatic.com/iframe.html?id=utils-clickaway-clickawayprovider--scrollable-click-away&viewMode=story
-   Vue: https://f0c5197d6--697a023f84e832e23544fb3c.chromatic.com/iframe.html?id=utils-clickaway-clickawayprovider--scrollable-click-away&viewMode=story

This build carries the story and none of the fix, so it shows the old behaviour.

To reproduce, click **Open level 1** inside the scrolling panel, then press and hold the grey
scrollbar on the panel's right edge.

Observe two things. The card disappears the moment the mouse button goes down, before you move the
mouse at all. The panel scrolls behind it, so the press did reach the scrollbar and did its job. Now
reopen the card and scroll the same panel with a wheel or a trackpad: the card stays open. That
contrast between the two ways of scrolling the same panel is the bug.

## What it breaks in lumapps-web

Open the page writer, then open the content settings panel. That panel has a fixed height and its own
scrollbar, and it holds a column of fields.

Open any dropdown in it, then reach for the panel's scrollbar to bring a field further down into
view. The dropdown closes before you have scrolled anywhere. Users lose the selection they were about
to make, and the only way to scroll while a dropdown is open is a wheel or a trackpad, which is not
discoverable. The bug reads as "dropdowns disappear during scroll".

Measured in the app under trusted input:

| measurement | value |
| --- | --- |
| `event.target` | the scrolling panel itself, not the dropdown or its anchor |
| `offsetX` against `clientWidth` | 393 against 385, so inside the 15px gutter |
| `scrollTop` when the dismissal fires | 0 |
| press on the track below the thumb | panel scrolls 0 to 195 **and** the dropdown closes |

The dropdown needs no relationship to the panel. A popover anchored elsewhere on the page closed too,
so this is not a nesting problem. The date picker closes the same way through a different popover
path, which is what rules out a fix in any single component.

## After the fix

-   React: https://495ba0e46--5fbfb1d508c0520021560f10.chromatic.com/iframe.html?id=utils-clickaway-clickawayprovider--scrollable-click-away&viewMode=story
-   Vue: https://495ba0e46--697a023f84e832e23544fb3c.chromatic.com/iframe.html?id=utils-clickaway-clickawayprovider--scrollable-click-away&viewMode=story

Same story, same presses. The card now stays open while you drag the scrollbar, and the panel still
scrolls. Pressing the panel content still closes the card.

# How to verify

Story: **utils / clickaway / ClickAwayProvider → Scrollable click away**.

Click **Open level 1** inside the panel, then:

| press | expected |
| --- | --- |
| scrollbar thumb, and drag it | stays open, panel scrolls |
| scrollbar track below the thumb | stays open, panel jumps down |
| wheel or trackpad over the panel | stays open |
| any line of panel content | closes |
| thick border of the middle block | closes |
| thick border of the truncated text block | closes |

The last two rows are the interesting controls. Both blocks overflow their box without ever rendering
a scrollbar, which is the case that made an earlier version of this guard swallow legitimate
dismissals.

The story forces the scrollbar to 15px with a `::-webkit-scrollbar` rule. Without it, a machine that
draws overlay scrollbars leaves no gutter to press and the story shows nothing. The same trap applies
to automated tests: a headless run without that rule passes on a broken build.

# Implementation notes

The new `isScrollbarPress` guard is geometric, never platform-based, and never calls
`preventDefault`, so dragging a scrollbar keeps working. Three conditions have to hold together:

-   the press lands outside the target's client box, measured with `clientLeft` and `clientTop` so a
    right-to-left layout with a left-hand scrollbar falls into the negative branches;
-   the axis overflows **and** its computed `overflow` actually renders a scrollbar, which excludes
    `hidden`, `clip`, and `visible`, so a bordered block of truncated text still dismisses;
-   coordinates are scaled back into layout space, because `getBoundingClientRect` is transformed
    while `clientWidth` is not.

The root element is measured against the viewport rather than against its own border box, which moves
with the scroll position. `visible` counts as scrollable there, because the root hands its overflow to
the viewport and still owns the window scrollbar.

`touchstart` is deliberately out of scope. A touch has no scrollbar to hit, and `TouchEvent` carries
no coordinates on the event itself, so the guard returns early for anything that is not a
`MouseEvent`.

Known limit: on an element that genuinely scrolls, its own border pixels beyond the client box read
as scrollbar. A 1px border has zero exposure, because the comparison is strict. The dead zone is
`borderWidth - 1` pixels, so only a chunky border on a scrollable container is affected.

Tests cover the geometry directly, since jsdom has no layout and every metric has to be stubbed: both
gutters, both axes, right-to-left, a scaled ancestor, the root element scrolled out of view, clipped
and visible overflow, a non-element target, and a target with no client box. The listener's decision
is covered through `setupClickAway`.

StoryBook lumx-react: https://495ba0e46--5fbfb1d508c0520021560f10.chromatic.com/
StoryBook lumx-vue: https://495ba0e46--697a023f84e832e23544fb3c.chromatic.com/

# Check list

-   [ ] Add `need: review-frontend` label
-   [ ] Add `need: test` label

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+-   `@lumx/core`, `@lumx/react`, `@lumx/vue`:
+    -   Click away: keep popovers open when a mouse press lands on a scrollbar. Pressing the scrollbar of a scrolling ancestor, or of the page itself, no longer counts as a click away. This affects every component that closes on click away, such as `Dropdown`, `Select`, `Combobox`, `PopoverDialog` and `Lightbox`.
+
 ## [4.20.0][] - 2026-07-24
 
 ### Added
--- a/packages/lumx-core/src/js/utils/ClickAway/TestStories.tsx
+++ b/packages/lumx-core/src/js/utils/ClickAway/TestStories.tsx
@@ -49,11 +49,8 @@ export function setup({
     };
 
     /**
-     * This story showcases how click away behaves around a scrollable container: scrolling a panel
-     * is not a click away, whichever way you scroll it, while pressing content is.
-     *
-     * The scrollbar is forced to take layout space, so that the story still offers a gutter to press
-     * on a system that draws overlay scrollbars.
+     * This story showcases that scrolling a container is not a click away, however you scroll it.
+     * The scrollbar takes layout space on purpose: overlay scrollbars leave no gutter to press.
      */
     const ScrollableClickAway = {
         render: () => (
--- a/packages/lumx-core/src/js/utils/ClickAway/TestStories.tsx
+++ b/packages/lumx-core/src/js/utils/ClickAway/TestStories.tsx
@@ -49,6 +49,65 @@ export function setup({
     };
 
     /**
+     * This story showcases how click away behaves around a scrollable container: scrolling a panel
+     * is not a click away, whichever way you scroll it, while pressing content is.
+     *
+     * The scrollbar is forced to take layout space, so that the story still offers a gutter to press
+     * on a system that draws overlay scrollbars.
+     */
+    const ScrollableClickAway = {
+        render: () => (
+            <>
+                <style>{`
+                    [data-demo-scroller]::-webkit-scrollbar { width: 15px; height: 15px; }
+                    [data-demo-scroller]::-webkit-scrollbar-track { background: #eee; }
+                    [data-demo-scroller]::-webkit-scrollbar-thumb { background: #888; }
+                `}</style>
+                <p>
+                    Scrolling a panel is not a click away. Open a level, then scroll the panel below with the wheel, or
+                    by dragging its grey scrollbar. The level stays open either way.
+                </p>
+                <p>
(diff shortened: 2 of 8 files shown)


gcornut:
Thanks for the investigation and the proposal
Although I think I'd prefer this fix https://github.com/lumapps/design-system/pull/1748, but I need to do some tests to check the viability

gcornut:
The initial bug seem to have another root cause. Closing this for now. We might revisit later

PR #1745, opened 2026-08-24
fix(slider): support drag on touch devices

lumbot:
# General summary

- `Slider` (`@lumx/react`) only wired drag via `mousedown`/`mousemove`/`mouseup` — touch devices only supported tap-to-select, not dragging.
- Added `onTouchStart` plus `touchmove`/`touchend`/`touchcancel` listeners mirroring the existing mouse-drag logic.
- Generalized position reading (`getPageX`/`isTouchEvent` helpers) so `getPercentValue`/`handleMove` work with both mouse and touch events.
- `touchmove` is registered with `{ passive: false }` and calls `preventDefault()` so page scroll doesn't hijack the drag.
- `touchcancel` is treated the same as `touchend` for listener cleanup, so an iOS system interrupt mid-drag doesn't leak listeners.
- No public API change — `onTouchStart` is internal only, `SliderProps` is unchanged.

# Screenshots

N/A — behavioral fix, no visual change.

LX-242

StoryBook lumx-react: https://85600456c--5fbfb1d508c0520021560f10.chromatic.com/

### DIFF
--- a/packages/lumx-react/src/components/slider/Slider.tsx
+++ b/packages/lumx-react/src/components/slider/Slider.tsx
@@ -85,6 +85,29 @@ const computePercentFromValue = (value: number, min: number, max: number): numbe
     Number((value - min) / (max - min));
 
 /**
+ * Type guard checking whether an event is a touch event (native or React synthetic).
+ *
+ * @param event The interaction event.
+ * @return Whether the event is a touch event.
+ */
+const isTouchEvent = (
+    event: React.MouseEvent | React.TouchEvent | MouseEvent | TouchEvent,
+): event is React.TouchEvent | TouchEvent => 'touches' in event;
+
+/**
+ * Read the horizontal page position from either a mouse or touch event.
+ *
+ * @param event The interaction event.
+ * @return The page X position.
+ */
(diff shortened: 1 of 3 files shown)


gcornut (changes requested):
Tested the SB chromatic deploy on iOS and this fix works fine

Some minor changes to do:
1. rename commit with prefix `fix(slider):`
2. add an entry in CHANGELOG > Unreleased > Fixed

packages/lumx-react/src/components/slider/Slider.tsx
```diff
@@ -245,6 +291,7 @@ export const Slider = forwardRef<SliderProps, HTMLDivElement>((props, ref) => {
                 }),
             )}
             onMouseDown={handleMouseDown}
+            onTouchStart={handleTouchStart}
```
gcornut, line 294: We need to add an entry in CHANGELOG > Unreleased > Fixed mentioning we fixed the slider on touch devices

gcornut (changes requested):

gcornut (approved):

gcornut:
Test OK

PR #1749, opened 2026-08-26
feat(link): underline links inside paragraphs

gcornut:
# General summary

First small step to improve `Link` accessibility in text: a link inside a `<p>` is now underlined by default, so it stays distinguishable from the surrounding text without relying on color alone ([WCAG 1.4.1 Use of Color](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html)).

-   `@lumx/core` — new SCSS rule in `components/link/_index.scss`. Compiles to `:where(p) .lumx-link`; the `:where()` keeps specificity at `0-1-0` (same as a bare `.lumx-link`) so consumers can override it without `!important`, from any stylesheet order.
-   **Scope is `<p>` only** for now — not `li`/`td`/`blockquote`, and not `Text as="span"`. Ship, measure, widen later.

# Screenshots

Left to CI — see the `In Paragraph` story in the deployed Storybooks and the visual-diff report.

StoryBook lumx-react: https://fb35713f3--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://fb35713f3--697a023f84e832e23544fb3c.chromatic.com/


packages/lumx-react/src/utils/react/wrapChildrenIconWithSpaces.tsx
```diff
@@ -18,9 +28,17 @@ export function wrapChildrenIconWithSpaces(children: React.ReactNode): React.Rea
...
         return child;
     });
 }
+
+/**
+ * Force wrap spaces around icons to make sure they are never stuck against text.
+ * Icons at the very start or end of the text are not padded on that side.
+ */
+export function wrapChildrenIconWithSpaces(children: React.ReactNode): React.ReactNode {
+    return wrap(children, true, true);
```
juanigalan91, line 53: I think this change also impacts `Text`'s icon spacing, not just `Link`'s. `Text.tsx` calls `wrapChildrenIconWithSpaces` directly too, and `Heading` renders through `Text`. So a `Text` or `Heading` starting or ending with an icon would lose its edge padding as well. Was that intended? If so, might be worth mentioning in the changelog too, since right now it only lists `Link`.

juanigalan91 (approved):

packages/lumx-react/src/utils/react/wrapChildrenIconWithSpaces.tsx
gcornut: Yes it's intended, I'll add it to the changelog

gcornut:
Fixing text wrapping when using non breakable space before icon in a text

gcornut:
Test OK ✅

PR #1755, opened 2026-09-03
fix(vue): remove selection chip on close icon click

gcornut:
# General summary

A click on the close icon of a chip did not remove the chip in the Vue `SelectTextField` in multiple mode. The React version works.

-   `@lumx/vue` `Chip`: `handleAfterClick` and `handleBeforeClick` called `stopPropagation()` in every case. They now do this only when a listener is registered. The React `Chip` already behaves this way through `useStopPropagation`.
-   Cause: the core `Chip` template always binds these two handlers. `SelectionChipGroup` removes a chip through one delegated `click` listener on the group container. The close icon lives in the chip `after` element, so the click never reached the container.
-   `@lumx/core`: new shared `SelectTextField` test that clicks the chip close icon. React and Vue both run this suite. The test fails without the fix.
-   Manual testing is not required. The new test covers the bug.

**Non-regression:** the change touches the Vue `Chip` `before` and `after` click handling. A consumer that registers `@afterClick` or `@beforeClick` keeps the same behavior, because propagation still stops in that case. All unit tests pass (1442).

# Screenshots

StoryBook lumx-react: https://a7749f228--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://a7749f228--697a023f84e832e23544fb3c.chromatic.com/


gcornut:
Test OK

juanigalan91 (approved):

PR #1767, opened 2026-09-11
fix(time-picker-field): snap out-of-bounds value to a selectable step option

lumbot:
## Summary

- `TimePickerField` (core/React/Vue): `snapTimeToBounds` now clamps out-of-bounds values to the nearest `step`-aligned option instead of the raw `minTime`/`maxTime`, so the field always lands on a visible/selectable entry.
- Both frameworks now pass `step` through to `snapTimeToBounds` at the `boundsMode: 'enforce'` effect/watch and the blur handler.
- When no `step`-aligned option exists inside `[minTime, maxTime]` (e.g. bounds too close together, or `minTime` pushes past end of day), the value is left unchanged rather than clamped to a non-existent option.
- Fixed a pre-existing test fixture (`maxTime: 18:15` with `step: 30`) that wasn't step-aligned, now updated to a grid-aligned bound.

LX-263

StoryBook lumx-react: https://1d1b10e74--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://1d1b10e74--697a023f84e832e23544fb3c.chromatic.com/


juanigalan91 (approved):

gcornut:
Test OK ✅

PR #1769, opened 2026-09-17
feat(input-required-legend): add InputRequiredLegend component (LX-260)

gcornut:
| Light | Dark |
 --- | ---
image | image

## Summary

- New `InputRequiredLegend` component (core/React/Vue) to explain the required field marker (`*`) at the top of a form. The wording is passed via the `label` prop (the design system holds no translation). It renders a `<p>` with a readable `*` marker followed by the text, styled like `InputHelper` (marker styled like `InputLabel`).
- `InputLabel`: the required marker is now hidden from assistive technologies via CSS alternative text (`content: "*" / ""`). The `required` attribute of the field already carries the state, so the marker only added a "star" announcement before each required label.
- Vue `TextField`: `isRequired` now sets the `required` attribute on the input (parity with React).
- `SelectTextField` / `Combobox.Input`: `isRequired` now sets `aria-required` on the input instead of the native `required` attribute. The input text is not the value of the field, so native constraint validation was wrong: in multiple selection the value is held by the chips and the input stays empty (a filled field was reported as empty), and with `filter="off"` the input is read only, so the native attribute did nothing. The label keeps its required marker. Consumers must validate the selection on submit.
- `TextField` (React & Vue): an explicit `required` attribute now wins over the one derived from `isRequired`. This is what lets `Combobox.Input` keep the label marker while replacing the attribute.

### Tests

Added to the shared core suites, so React and Vue both run them:
- `SelectTextField/Tests.tsx`: `aria-required` is set, the native attribute is absent, `validity.valueMissing` stays false (single and multiple selection), the label is marked required, and nothing is set by default.
- `TextField/Tests.ts`: an explicit `required` attribute overrides `isRequired` while the label marker stays.

LX-260

StoryBook lumx-react: https://82f86e7f6--5fbfb1d508c0520021560f10.chromatic.com/

StoryBook lumx-vue: https://82f86e7f6--697a023f84e832e23544fb3c.chromatic.com/


juanigalan91 (approved):

packages/lumx-core/src/js/components/SelectTextField/Tests.tsx
```diff
@@ -890,6 +892,50 @@ export default function selectTextFieldTests({ components, renderWithState }: Se
...
+        it('should set aria-required and no native required in multiple selection', () => {
+            renderWithState(multiTemplate, { isRequired: true, value: [FRUITS[0]] });
+            const input = getByRole(document.body, 'combobox') as HTMLInputElement;
+
+            expect(input.getAttribute('aria-required')).toBe('true');
+            expect(input.hasAttribute('required')).toBe(false);
+            expect(input.validity.valueMissing).toBe(false);
+        });
+
+        it('should not set aria-required by default', () => {
```
juanigalan91, line 942: Could we add a case for `filter="off"`? That's the read-only scenario this fix is for, and nothing in this suite exercises `filter="off"` at all right now.

packages/lumx-core/src/scss/components/input-label/_index.scss
```diff
@@ -6,7 +6,10 @@
     &--is-required::before {
         margin-right: $lumx-spacing-unit-tiny;
         font-weight: 700;
-        content: "*";
+
+        // Empty alternative text: hides the marker from assistive technologies
+        // (the `required` attribute of the field already carries the state)
+        content: "*" / "";
```
juanigalan91, line 12: Select and SelectMultiple also use `isRequired` on the label, but they're a div, not a form control, and never set `aria-required`. With the marker now hidden from assistive tech, `isRequired` on those two gives no signal at all. Worth adding `aria-required` there too, or is that out of scope since they're deprecated?

gcornut: Out of scope here. `Select` and `SelectMultiple` are both `@deprecated use SelectTextField instead`, and `SelectTextField` does set `aria-required` in this PR.

One note for whoever picks up the deprecated pair: `aria-required` alone would not fix them. Their trigger is a `div` with no role, and `aria-required` is valid on a small set of roles only (`combobox`, `listbox`, `textbox`, …). A screen reader ignores it there. A real fix needs a role on the trigger first, which is a full a11y rework.

gcornut:
Test OK

